PR #15, opened 2017-05-19
Travis submodule fix

shaun-edwards:
Partially reverts changes in #13.  This should now work as the submodule in `prettyprint` has been replaced.


shaun-edwards:
It appears the travis issue is related to `industrial_ci` building in the install space (it didn't always do this).  See [here](https://github.com/ros-industrial/industrial_ci/issues/176) for more info.  This must be fixed in the `prettyprint` repo.

The CI now fails because of unit test dependencies (forward progress!)

PR #17, opened 2017-05-30
Splits Execute Method into 'plan' and 'execute' methods

geoffreychiou:
@shaun-edwards Is this what you were thinking?

### DIFF
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -51,7 +51,7 @@ class JointTrajectoryPoint;
 class CartTrajectoryPoint;
 class Robot;
 typedef std::vector<std::unique_ptr<TrajectoryPoint> > Trajectory;
-
+typedef control_msgs::FollowJointTrajectoryGoal TrajectoryGoal;
 
 /**
  * @brief Robot is a wrapper around standard MoveIt objects.  It makes multiple
@@ -142,12 +142,23 @@ public:
                         std::vector<double> & joint_point) const;
 
   /**
-   * @brief execute a given trajectory
+   * @brief plans a given trajectory
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
-   * @return - true if desired trajectory was executed.
+   * @param goal - control_msgs::FollowJointTrajectoryGoal
+   * @return - true if desired trajectory was planned
    */
-  bool execute(const std::string traj_name);
+  bool plan(const std::string traj_name, TrajectoryGoal &goal);
+
+  /**
+   * @brief executes a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param goal - control_msgs::FollowJointTrajectoryGoal (must have been planned)
+   * @return - true if desired trajectory was executed
+   */
+  bool execute(const std::string traj_name, const TrajectoryGoal &goal);
+
   /**
    * @brief clearTrajectory - clears stored trajectory
    * @param traj_name - trajectory to clear
--- a/moveit_simple/src/moveit_simple.cpp
+++ b/moveit_simple/src/moveit_simple.cpp
@@ -352,46 +352,6 @@ bool Robot::execute(const std::string traj_name,
   return success;
 }
 
-bool Robot::execute(const std::string traj_name)
-{
-  std::lock_guard<std::recursive_mutex> guard(m_);
-
-  const double TIMEOUT_SCALE = 1.25;  //scales time to wait for action timeout.
-  bool success = false;
-  if ( traj_map_.count(traj_name) )
-  {
-    control_msgs::FollowJointTrajectoryGoal goal;
-    goal.trajectory.joint_names = joint_group_->getVariableNames();
-    if ( toJointTrajectory(traj_name, goal.trajectory.points) )
-    {
-      ros::Duration traj_time =
-          goal.trajectory.points[goal.trajectory.points.size()-1].time_from_start;
-      ros::Duration timeout(TIMEOUT_SCALE * traj_time.toSec());
-      if (action_.sendGoalAndWait(goal, timeout) == actionlib::SimpleClientGoalState::SUCCEEDED)
-      {
-        ROS_INFO_STREAM("Successfully executed trajectory: " << traj_name);
-        success = true;
-      }
-      else
-      {
-        ROS_WARN_STREAM("Trajectory " << traj_name << " failed to exectue");
-        success = false;
-      }
-    }
-    else
-    {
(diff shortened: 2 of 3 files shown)


moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -142,12 +142,23 @@ class Robot
...
   /**
-   * @brief execute a given trajectory
+   * @brief plans a given trajectory
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
-   * @return - true if desired trajectory was executed.
+   * @param goal - control_msgs::FollowJointTrajectoryGoal
+   * @return - true if desired trajectory was planned
    */
-  bool execute(const std::string traj_name);
```
shaun-edwards, line 150: Normally I would have left the original method in (with a deprecated notice).  That way you don't break others code.  Don't worry about it for now.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -300,49 +300,59 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
-      ROS_ERROR_STREAM("Failed to convert " << traj_name << " to joint trajectory");
+      ROS_WARN_STREAM("Failed to convert " << traj_name << " to joint trajectory");
       success = false;
     }
+  }
+  else {success = false;}
+  return success;
+}
 
+bool Robot::execute(const std::string traj_name,
```
shaun-edwards, line 326: Why is the `traj_name` argument required in this method?  It looks like it is only used for logging, whereas in the previous implementation it was required to look-up the trajectory.

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: Actually, after seeing your tests, I like usability of this method.  The "new" methods give you granular control for testing and other uses, but we still need this method for simplicity.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -62,7 +62,21 @@ TEST(MoveitSimpleTest, add_trajectory)
   EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint2", 3.0));
   EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint3", 4.0));
 
-  EXPECT_TRUE(robot.execute(TRAJECTORY_NAME));
+  // Check if a proper plan and execution works.
+  moveit_simple::TrajectoryGoal trajectory_goal;
+  EXPECT_TRUE(robot.plan(TRAJECTORY_NAME, trajectory_goal));
```
shaun-edwards, line 67: This test shows how we've made things harder to use...see comments above.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -62,7 +62,21 @@ TEST(MoveitSimpleTest, add_trajectory)
...
+  // Check if a trajectory without any trajPoints is passed in plan
+  moveit_simple::TrajectoryGoal random_goal;
+  EXPECT_FALSE(robot.plan("Random Name", random_goal));
+
+  // Check if a trajectory with an empty goal is passed in to execution
+  moveit_simple::TrajectoryGoal empty_goal;
+  EXPECT_FALSE(robot.execute(TRAJECTORY_NAME, empty_goal));
+
+  // Check if a wrong name is passed into execution
+  EXPECT_FALSE(robot.execute("Random Name", empty_goal));
```
shaun-edwards: How does this work...where does execute verify that the trajectory name is correct?

moveit_simple/test/moveit_simple.cpp
```diff
@@ -62,7 +62,21 @@ TEST(MoveitSimpleTest, add_trajectory)
...
+  EXPECT_TRUE(robot.plan(TRAJECTORY_NAME, trajectory_goal));
+  EXPECT_TRUE(robot.execute(TRAJECTORY_NAME, trajectory_goal));
+
+  // Check if a trajectory without any trajPoints is passed in plan
+  moveit_simple::TrajectoryGoal random_goal;
+  EXPECT_FALSE(robot.plan("Random Name", random_goal));
+
+  // Check if a trajectory with an empty goal is passed in to execution
+  moveit_simple::TrajectoryGoal empty_goal;
+  EXPECT_FALSE(robot.execute(TRAJECTORY_NAME, empty_goal));
```
shaun-edwards, line 76: Good check!

shaun-edwards (changes requested):

moveit_simple/test/moveit_simple.cpp
geoffreychiou: Good point, this is a bad check, I'll remove it. It was only passing because the test because it failed due to an empty goal.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Should we not pass it in even if it is only used for logging?

moveit_simple/test/moveit_simple.cpp
geoffreychiou: Would something like this be easier?
``` C++
struct TrajectoryDefinition
{
  TrajectoryDefinition(const std::string &input_name) : name(input_name)
  moveit_simple::TrajectoryGoal goal;
  const std::string name;
};

moveit_simple::TrajectoryDefinition trajectory("traj1");
robot.plan(trajectory);
robot.execute(trajectory);
```

moveit_simple/src/moveit_simple.cpp
shaun-edwards: This relies on the user of you class/method to do the right thing.  The argument also impies it's used in some meaningful way, but that's not true.

moveit_simple/test/moveit_simple.cpp
shaun-edwards: I'm not sure how defining a new structure makes it easier.  My main concern was getting rid of a method that planned and executed a trajectory in one shot.

shaun-edwards:
@geoffreychiou, don't forget about this one.

moveit_simple/test/moveit_simple.cpp
nsnitish: @shaun-edwards ! Won't it be easier for user to execute the trajectory by just passing the traj_name? trajectory_goal can be passed internally by using a simple map

 `std::map<std::string, moveit_simple::TrajectoryGoal> planned_traj_map_`

shaun-edwards: Or we could support both and/or all combinations of plan/execute.

nsnitish: Is it like
1. `planAndExecute(traj_name)`
2. `plan(traj_name, trajectory_goal)` and `execute(trajectory_goal)` ?

PR #18, opened 2017-07-05
Exception handling for addTrajPoint and execute methods

nsnitish:


nsnitish:
@shaun-edwards! Can you please have a look?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -102,35 +102,37 @@ bool Robot::addTrajPoint(const std::string & traj_name, const Eigen::Affine3d po
...
     traj_map_[traj_name].push_back(std::move(point));
-    success = true;
+//    success = true;
   }
   else
   {
+
     ROS_ERROR_STREAM("Failed to add point for trajectory " << traj_name);
-    success = false;
+    throw std::invalid_argument("shit! a nullptr");
```
shaun-edwards: Watch your language...git saves every version, so now this will be in there for good.  Please reset your commits, correct, this and then do a force push to clean this out from history.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -52,17 +52,19 @@ TEST(MoveitSimpleTest, add_trajectory)
...
   ROS_INFO_STREAM("Testing trajectory adding of points");
-  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5));
-  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint1", 1.0));
-  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "tf_pub1",   2.0));
-  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint2", 3.0));
-  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint3", 4.0));
-
-  EXPECT_TRUE(robot.execute(TRAJECTORY_NAME));
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5));
+//  EXPECT_TRUE(robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5));
```
shaun-edwards: These should bu updated to EXPECT_NO_THROW.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -52,17 +52,19 @@ TEST(MoveitSimpleTest, add_trajectory)
   const std::string TRAJECTORY_NAME("traj1");
 
   ROS_INFO_STREAM("Testing loading of unknown point, should fail");
-  EXPECT_FALSE(robot.addTrajPoint("bad_traj", "unknown_name", 1.0));
+//  EXPECT_FALSE(robot.addTrajPoint("bad_traj", "unknown_name", 1.0));
+  EXPECT_ANY_THROW(robot.addTrajPoint("bad_traj", "unknown_name", 1.0));
```
shaun-edwards: Have you verified this test indeed throws.  I don't think your changes above will result in this throwing.

shaun-edwards: Maybe it does.  I think I see it now.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -88,7 +88,7 @@ bool Robot::addTrajPoint(const std::string & traj_name, const Eigen::Affine3d po
     Eigen::Affine3d pose_rel_robot = transformToBase(pose, frame);
     std::unique_ptr<TrajectoryPoint> point =
       std::unique_ptr<TrajectoryPoint>(new CartTrajectoryPoint(pose_rel_robot, time, point_name));
-    success = addTrajPoint(traj_name, point);
+//    success = addTrajPoint(traj_name, point);
```
shaun-edwards: You should still need to call `addTrajPoint`, I don't think you can just delete it.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -88,7 +88,7 @@ bool Robot::addTrajPoint(const std::string & traj_name, const Eigen::Affine3d po
...
   }
   catch (tf2::TransformException &ex)
```
shaun-edwards: I recommend catching this error, logging the problem, and then re-throwing the exception to a higher level or creating a class of exception that all `addTrajPoint` methods call.

shaun-edwards (changes requested):

moveit_simple/src/moveit_simple.cpp
nsnitish: I haven't yet changed this part of the code. It was done to see if other calls work.

moveit_simple/src/moveit_simple.cpp
nsnitish: We are already throwing an exception in [transformToBase](https://github.com/shaun-edwards/moveit_simple/blob/indigo-devel/moveit_simple/src/moveit_simple.cpp#L457). Do we need to catch and re-throw here as well?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -74,63 +74,67 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
 {
   std::lock_guard<std::recursive_mutex> guard(m_);
 
-  bool success = false;
+//  bool success = false;
   ROS_INFO_STREAM("Attempting to add " << point_name << " to " << traj_name << "relative to"
                   << frame << " at time " << time);
-  try
-  {
+//  try
```
shaun-edwards: Please remove code instead of commenting it out.  It makes it easier to review.  You can always revert back.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -74,63 +74,67 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
     traj_map_[traj_name].push_back(std::move(point));
-    success = true;
+//    success = true;
   }
   else
   {
+
     ROS_ERROR_STREAM("Failed to add point for trajectory " << traj_name);
-    success = false;
+    throw std::runtime_error("not a valid point to add");
```
shaun-edwards: You should include the arguments to the function to make it easier for the developer to track down.  Say something like "Null point cannot be added to <traj_name>"

moveit_simple/src/moveit_simple.cpp
```diff
@@ -174,14 +178,16 @@ std::unique_ptr<TrajectoryPoint> Robot::lookupTrajectoryPoint(const std::string
...
+      throw ex;
     }
   }
 
   else
   {
     ROS_ERROR_STREAM("Failed to find point " << name << ", consider implementing more look ups");
-    return std::unique_ptr<TrajectoryPoint>(nullptr);
+//    return std::unique_ptr<TrajectoryPoint>(nullptr);
+    throw std::runtime_error("No point found");
```
shaun-edwards: Add point name to the exception.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -230,10 +236,30 @@ bool Robot::isInCollision(const Eigen::Affine3d pose, const std::string & frame,
...
+// {
+//   ROS_WARN_STREAM("Reacheabilioty failed for arbitrary pose point: " << ex.what());
+//   reacheable = false;
+// }
+// catch(std::runtime_error &re)
+// {
+//   ROS_WARN_STREAM("Reacheability failed as : " << ex.what());
+//   reacheable = false;
+// }
+  return reacheable;
```
shaun-edwards: It's OK to return at the end of the try block.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -320,27 +346,30 @@ bool Robot::execute(const std::string traj_name)
...
         ROS_INFO_STREAM("Successfully executed trajectory: " << traj_name);
-        success = true;
+//        success = true;
       }
       else
       {
         ROS_WARN_STREAM("Trajectory " << traj_name << " failed to exectue");
-        success = false;
+//        success = false;
+        throw std::runtime_error("Execution Failure");
```
shaun-edwards: Better string description (see comments above).

moveit_simple/test/moveit_simple.cpp
```diff
@@ -32,9 +32,13 @@ TEST(MoveitSimpleTest, construction)
 TEST(MoveitSimpleTest, reachability)
 {
   moveit_simple::Robot robot(ros::NodeHandle(), "robot_description", "manipulator");
+  const Eigen::Affine3d pose = Eigen::Affine3d::Identity(); 
 
   ROS_INFO_STREAM("Testing reachability of unknown point, should fail");
   EXPECT_FALSE(robot.isReachable("unknown_name"));
+  EXPECT_FALSE(robot.isReachable(pose, "random_link"));
```
shaun-edwards, line 39: Thanks for adding tests!

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -111,7 +111,7 @@ class Robot
    * @param time - time from start of trajectory to reach point
    * @return true if point successfullly added
    */
-  bool addTrajPoint(const std::string & traj_name, const std::string & point_name,
+  void addTrajPoint(const std::string & traj_name, const std::string & point_name,
```
shaun-edwards: Please update header comments with the possible exceptions.  I believe the exceptions are added as stated [here](http://cs.swan.ac.uk/~csoliver/ok-sat-library/internet_html/doc/doc/Doxygen/1.7.6.1/html/commands.html#cmdexception)

shaun-edwards (changes requested):
Looking good.  Just a few thoughts.  

Do you think you will create special exceptions for this class or stick with the standard exceptions?

moveit_simple/src/moveit_simple.cpp
nsnitish: But we also have to return false if we catch the exception.

nsnitish:
I think sticking to standard exceptions will be a good idea for this case.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -145,9 +144,11 @@ class Robot
    * @brief execute a given trajectory
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
-   * @return - true if desired trajectory was executed.
+   * \throws <std::runtime_error> (Execution failure)
```
shaun-edwards: This method throws for multiple reasons.  Because it always throws the `std::runtime_error`, the calling code will not be able to handle these error cases separately.  Isn't this a problem?

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: I think it can be handled like this.

try{
// do something
}
catch(std::exception& ex)
{
std::cout << "Exception: " << ex.what() << "\n";
// divide into multiple cases how a particular exception(ex.what()) should be handled.
}

shaun-edwards: Are you suggesting, we programatically parse the error string to determine how we should respond?  The `ex.what()` is useful for logging and display, but it's not generally a good idea to use it for recovering errors.  You left that part out of your code snippet, but try and stub it out.  It's complex and fragile at best.

nsnitish: Ok. I thought the exceptions here will be used for display and debug purposes and at best, we might need to log them in different files. Now that you have mentioned this, I think defining class for atleast execute method will make more sense. Thanks for pointing that out.

nsnitish:
@shaun-edwards! Does this look like a better implementation?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +360,47 @@ class CartTrajectoryPoint : public TrajectoryPoint
   Eigen::Affine3d pose_;
 };
 
+class addTrajPointException: public std::runtime_error
```
shaun-edwards: Can this be made more simple?  Exceptions should indicate root cause and not be tied to code structure.  Creating a hierarchy of exceptions for such a simple class is another indicator that it's too complex.  Think about this from the user/developer point of view.  I called a specific method or set of methods and it failed?  Why?  Here are the reasons I can think of:
 - invalid name (trajectory or point)
 - bad conversion or perhaps more specifically we couldn't convert a cartesian point to a joint point due to an IK failure, so maybe this is an IK failure
 - move execution fail (something in the sendAndWait method failed)

Can you think of any other root causes?  The number of exception classes should reflect the number of root causes.  All exceptions should inherit from `std::runtime_error`.  More complex inheritance (not just in the case of exceptions, but all code) is almost never needed.

shaun-edwards (changes requested):
I think we've gone to something that is too complex.  Let's see if we can make it more simple.

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: Are you suggesting that exceptions should be defined for the whole namespace and defining them hierarchically for each method is not a good idea?

shaun-edwards: Yes.  Take the standard library for example.  The exception hierarchy is defined outside of any of the standard libraries/templates.  It's because failure modes are not unique to any function/class, but rather are cross cutting.  In this case, I expect it will result in far fewer exceptions.  It's easier to code, easier to document, and easier for others to use.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -74,63 +74,50 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
   if( point )
   {
     traj_map_[traj_name].push_back(std::move(point));
-    success = true;
   }
   else
   {
     ROS_ERROR_STREAM("Failed to add point for trajectory " << traj_name);
-    success = false;
+    throw nullPointException("Null point can't be added to "+ traj_name);
```
shaun-edwards: This should be an [invalid argument exception](http://www.cplusplus.com/reference/stdexcept/invalid_argument/)

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -51,6 +51,10 @@ class JointTrajectoryPoint;
 class CartTrajectoryPoint;
 class Robot;
 typedef std::vector<std::unique_ptr<TrajectoryPoint> > Trajectory;
+class nullPointException;
```
shaun-edwards: By ROS convention, classes should start with a capital letter

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -109,9 +113,9 @@ class Robot
    * @param traj_name - name of trajectory buffer to add point to
    * @param point_name - name of point to add
    * @param time - time from start of trajectory to reach point
-   * @return true if point successfullly added
+   * \throws <std::runtime_error> (If point_name is not found)
```
shaun-edwards: Be sure to update the header comments once we have settled on an exception set.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -74,63 +74,50 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
 
-bool Robot::addTrajPoint(const std::string & traj_name, const std::string & point_name,
+void Robot::addTrajPoint(const std::string & traj_name, const std::string & point_name,
                          double time)
 {
   std::lock_guard<std::recursive_mutex> guard(m_);
 
   ROS_INFO_STREAM("Attempting to add " << point_name << " to " << traj_name
                  << " at time " << time);
   std::unique_ptr<TrajectoryPoint> point = lookupTrajectoryPoint(point_name, time);
```
shaun-edwards: You should throw an exception from here.  This is the point at which you know why things failed.  The `point_name` is not valid.  The null pointer exception that would get thrown in the next function call isn't useful if you don't know why a null pointer was passed.  In fact it's confusing, because the user never passed a pointer to the method.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +357,29 @@ class CartTrajectoryPoint : public TrajectoryPoint
   Eigen::Affine3d pose_;
 };
 
+class nullPointException: public std::runtime_error
```
shaun-edwards: In my previous review, I identified 3 failures, which should have mapped to 3 exceptions.  There are 4.  Is there a reason you added one.  The `nullPointException` and the `noTrajNameException` are the same general error.  A `name` was passed to a method that isn't recognized.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,19 +54,24 @@ TEST(MoveitSimpleTest, add_trajectory)
 {
```
shaun-edwards, line 54: Please add additional tests to cover all possible exceptions.

shaun-edwards (changes requested):

moveit_simple/src/moveit_simple.cpp
```diff
@@ -74,63 +74,50 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
-  return addTrajPoint(traj_name, point);
+  addTrajPoint(traj_name, point);
 }
 
 
 
 
-bool Robot::addTrajPoint(const std::string & traj_name,
+void Robot::addTrajPoint(const std::string & traj_name,
```
nsnitish: @shaun-edwards! Is this method explicitly required? Only call of interest is
traj_map_[traj_name].push_back(std::move(point));
It is already made sure that it is a valid point or an exception can be thrown. It is more confusing to throw something here.
P.S. it is an internal method

moveit_simple/src/moveit_simple.cpp
```diff
@@ -174,14 +164,14 @@ std::unique_ptr<TrajectoryPoint> Robot::lookupTrajectoryPoint(const std::string
     catch (tf2::TransformException &ex)
     {
       ROS_ERROR_STREAM("TF transform lookup failed: " << ex.what());
-      return std::unique_ptr<TrajectoryPoint>(nullptr);
+      throw ex;
```
nsnitish, line 167: Not sure how to write a test for this exception.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,19 +54,33 @@ TEST(MoveitSimpleTest, add_trajectory)
...
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "tf_pub1",   2.0));
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint2", 3.0));
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint3", 4.0));
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, pose, "tool0", 5.0, "random_name"));
+  EXPECT_NO_THROW(robot.execute(TRAJECTORY_NAME));
+
+  EXPECT_NO_THROW(robot.addTrajPoint("traj2", "waypoint4", 4.5));
+  EXPECT_THROW(robot.execute("traj2"), moveit_simple::IKFailException);
+
+// Not sure if it is a good test for Execution Failure
```
shaun-edwards: Let's take this test out.  It really just takes advantage of a bug.  If you can't think of another way to test, that OK.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -121,10 +124,9 @@ class Robot
    * @param frame - frame (must be a TF accessible frame) in which pose is defined
    * @param time - time from start of trajectory to reach point
    * @param point_name - (optional) name of point (used in log messages)
-   *
-   * @return true if point successfullly added
+   * \throws <tf2::TransformException> (Transform from frame to robot base failed)
```
shaun-edwards: Does `@throws` works too?  If so, let's stick with the convention of this file (here and comments below).

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +356,23 @@ class CartTrajectoryPoint : public TrajectoryPoint
   Eigen::Affine3d pose_;
 };
 
+
+
+
+class ExecutionFailureException: public std::runtime_error
```
shaun-edwards: Add doxygen comments

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +356,23 @@ class CartTrajectoryPoint : public TrajectoryPoint
...
+{ 
+public:
+  ExecutionFailureException(const std::string errorDescription) : std::runtime_error(errorDescription) { ; };
```
shaun-edwards, line 372: Add doxygen comments

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +356,23 @@ class CartTrajectoryPoint : public TrajectoryPoint
...
+};
+
+
+
+
+class IKFailException: public std::runtime_error
```
shaun-edwards: Add doxygen comments

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -352,4 +356,23 @@ class CartTrajectoryPoint : public TrajectoryPoint
...
+{ 
+public:
+  IKFailException(const std::string errorDescription) : std::runtime_error(errorDescription) { ; };
```
shaun-edwards, line 387: Add doxygen comments

shaun-edwards (changes requested):
Just some simple changes.  After that looks good to merge.  Thanks!

shaun-edwards (approved):
Thanks for the contribution!

PR #19, opened 2017-07-15
Cartesian Interpolation in Trajectory

nsnitish:
Added toCartTrajPoint function

### DIFF
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -144,12 +144,12 @@ public:
                         std::vector<double> & joint_point) const;
 
   /**
-   * @brief getFK finds cartesian pose for the given joint positions.
+   * @brief getPosition finds cartesian pose for the given joint positions.
    * @param joint_point - joint positions
    * @param pose - pose corresponding to joint_pose
    * @return true if pose is found
    */
-  bool getFK(const std::vector<double> & joint_point,
+  bool getPosition(const std::vector<double> & joint_point,
              Eigen::Affine3d & pose) const;
 
   /**
@@ -198,6 +198,9 @@ protected:
               unsigned int attempts=1) const;
   bool getIK(const Eigen::Affine3d pose, std::vector<double> & joint_point,
              double timeout=1, unsigned int attempts=1) const;
+
+  bool getFK(const std::vector<double> & joint_point,
+                               Eigen::Affine3d &pose) const;
 
   std::unique_ptr<TrajectoryPoint> lookupTrajectoryPoint(const std::string & name,
                                                               double time) const;
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -130,12 +130,12 @@ public:
    * @param type - Type of interpolation from last point to this point
    * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
    * @param num_steps - number of points to be interpolated
-   * By deafult, it is set to 0
+   * By deafult, it is set to 0 and only adds the given point to the Trajectory
    * @throws <std::invalid_argument> (point_name is not found)
    * @throws <tf2::TransformException> (transform of TF named point_name fails)
    */
   void addTrajPoint(const std::string & traj_name, const std::string & point_name,
-                    double time, const InterpolationType & type = JOINT,
+                    double time, const InterpolationType & type = interpolation_type::JOINT,
                     const unsigned int num_steps = 0);
   /**
    * @brief Add trajectory point to motion buffer
@@ -147,13 +147,13 @@ public:
    * @param type - Type of interpolation from last point to this point
    * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
    * @param num_steps - number of points to be interpolated
-   * By deafult, it is set to 0
+   * By deafult, it is set to 0 and only adds the given point to the Trajectory
    * @param point_name - (optional) name of point (used in log messages)
    * @throws <tf2::TransformException> (Transform from frame to robot base failed)
   */
   void addTrajPoint(const std::string & traj_name, const Eigen::Affine3d pose,
                     const std::string & frame, double time,
-                    const InterpolationType & type = JOINT,
+                    const InterpolationType & type = interpolation_type::JOINT,
                     const unsigned int num_steps = 0,
                     const std::string & point_name = std::string());
 
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -144,6 +144,15 @@ public:
                         std::vector<double> & joint_point) const;
 
   /**
+   * @brief getFK finds cartesian pose for the given joint positions.
+   * @param joint_point - joint positions
+   * @param pose - pose corresponding to joint_pose
+   * @return true if pose is found
+   */
+  bool getFK(const std::vector<double> & joint_point,
+             Eigen::Affine3d & pose) const;
+
+  /**
    * @brief execute a given trajectory
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
@@ -271,6 +280,9 @@ protected:
   virtual std::unique_ptr<JointTrajectoryPoint> toJointTrajPoint(
       const Robot & robot,  double timeout, const std::vector<double> & seed) const=0;
 
+  virtual std::unique_ptr<CartTrajectoryPoint> toCartTrajPoint(
+      const Robot & robot) const=0;
+
   double t_;
   std::string name_;
   PointType type_;
@@ -302,11 +314,10 @@ public:
 protected:
 
   virtual std::unique_ptr<JointTrajectoryPoint> toJointTrajPoint(
-      const Robot & robot,  double timeout, const std::vector<double> & seed) const
-  {
-    ROS_DEBUG_STREAM("JointTrajectoryPoint: passing through joint trajectory point");
-    return std::unique_ptr<JointTrajectoryPoint>(new JointTrajectoryPoint(*this));
-  }
+      const Robot & robot,  double timeout, const std::vector<double> & seed) const;
+
+  virtual std::unique_ptr<CartTrajectoryPoint> toCartTrajPoint(
+      const Robot & robot) const;
 
 private:
   std::vector<double> joint_point_;
@@ -336,21 +347,10 @@ public:
 protected:
 
   virtual std::unique_ptr<JointTrajectoryPoint> toJointTrajPoint(
-      const Robot & robot,  double timeout, const std::vector<double> & seed) const
-  {
-    std::vector<double> joints;
+      const Robot & robot,  double timeout, const std::vector<double> & seed) const;
 
-    ROS_DEBUG_STREAM("CartTrajectoryPoint: Calculating IK for joint trajectory point");
-    if( robot.getJointSolution(pose_, timeout, seed, joints) )
-    {
-      return std::unique_ptr<JointTrajectoryPoint>( new JointTrajectoryPoint(joints, time(), name()));
-    }
-    else
-    {
-      ROS_WARN_STREAM("Failed to find joint solution for point: " << name_);
-      return std::unique_ptr<JointTrajectoryPoint>(nullptr);
-    }
-  }
+  virtual std::unique_ptr<CartTrajectoryPoint> toCartTrajPoint(
+      const Robot & robot) const;
 
 private:
   Eigen::Affine3d pose_;
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -213,7 +213,7 @@ public:
    */
   Eigen::Affine3d interpolate( const Eigen::Affine3d & from,
                                const Eigen::Affine3d & to,
-                               double t);
+                               double t) const;
 
   /**
    * @brief interpolate - returns the joint point interpolated from \e from joint point
@@ -224,7 +224,7 @@ public:
    */
   std::vector<double> interpolate( const std::vector<double> & from,
                                   const std::vector<double> & to,
-                                  double t);
+                                  double t) const;
 
 protected:
   Robot();
@@ -270,7 +270,8 @@ protected:
    */
   void interpolate( const std::unique_ptr<CartTrajectoryPoint>& from,
                     const std::unique_ptr<CartTrajectoryPoint>& to,
-                    double t, std::unique_ptr<CartTrajectoryPoint> & point);
+                    double t, 
+                    std::unique_ptr<CartTrajectoryPoint> & point) const;
 
   /**
    * @brief interpolate - Joint interpolation from \e from point towards \e to
@@ -282,10 +283,13 @@ protected:
    */
   void interpolate( const std::unique_ptr<JointTrajectoryPoint>& from,
                     const std::unique_ptr<JointTrajectoryPoint>& to,
-                    double t, std::unique_ptr<JointTrajectoryPoint> & point);
+                    double t, 
+                    std::unique_ptr<JointTrajectoryPoint> & point) const;
 
   void addTrajPoint(const std::string & traj_name,
-                    std::unique_ptr<TrajectoryPoint> &point);
+                    std::unique_ptr<TrajectoryPoint> &point,
+                    const InterpolationType & type = interpolation_type::JOINT,
+                    const unsigned int num_steps = 0);
 
   bool isReachable(std::unique_ptr<TrajectoryPoint> & point, double timeout,
                    std::vector<double> joint_seed = std::vector<double>() ) const;
--- a/moveit_simple/include/moveit_simple/moveit_simple.h
+++ b/moveit_simple/include/moveit_simple/moveit_simple.h
@@ -56,6 +56,19 @@ class ExecutionFailureException;
 class IKFailException;
 
 
+enum InterpolationType
+   {
+      UNKNOWN = 0,
+      JOINT = 1,
+      CARTESIAN = 2
+   };
+struct TrajectoryPointInfo {
+  std::unique_ptr<TrajectoryPoint> point;
+  InterpolationType type;
+  unsigned int num_steps;
+};
+typedef std::vector<TrajectoryPointInfo> TrajectoryInfo;
+
 /**
  * @brief Robot is a wrapper around standard MoveIt objects.  It makes multiple
  assumptions about the type and complexity of the robot.  Assumptions are:
@@ -114,11 +127,13 @@ public:
    * @param time - time from start of trajectory to reach point
    * @param traj_type - Type of Trajectory from last point to this point
    * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
+   * @param num_steps - number of points to be interpolated
+   * By deafult, it is set to 0
    * @throws <std::invalid_argument> (point_name is not found)
    * @throws <tf2::TransformException> (transform of TF named point_name fails)
    */
   void addTrajPoint(const std::string & traj_name, const std::string & point_name,
-                    double time, const std::string & traj_type = "JOINT",
+                    double time, const InterpolationType & type = JOINT,
                     const unsigned int num_steps = 0);
   /**
    * @brief Add trajectory point to motion buffer
@@ -129,12 +144,14 @@ public:
    * @param time - time from start of trajectory to reach point
    * @param traj_type - Type of Trajectory from last point to this point
    * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
+   * @param num_steps - number of points to be interpolated
+   * By deafult, it is set to 0
    * @param point_name - (optional) name of point (used in log messages)
    * @throws <tf2::TransformException> (Transform from frame to robot base failed)
   */
   void addTrajPoint(const std::string & traj_name, const Eigen::Affine3d pose,
                     const std::string & frame, double time,
-                    const std::string & traj_type = "JOINT",
+                    const InterpolationType & type = JOINT,
                     const unsigned int num_steps = 0,
                     const std::string & point_name = std::string());
 
@@ -152,12 +169,12 @@ public:
                         std::vector<double> & joint_point) const;
 
   /**
-   * @brief getPosition finds cartesian pose for the given joint positions.
+   * @brief getPose finds cartesian pose for the given joint positions.
    * @param joint_point - joint positions
    * @param pose - pose corresponding to joint_pose
    * @return true if pose is found
    */
-  bool getPosition(const std::vector<double> & joint_point,
+  bool getPose(const std::vector<double> & joint_point,
              Eigen::Affine3d & pose) const;
 
   /**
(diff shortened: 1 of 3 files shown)


moveit_simple/src/moveit_simple.cpp
```diff
@@ -478,6 +478,24 @@ Eigen::Affine3d Robot::transformToBase(const Eigen::Affine3d &in,
 
 
 
+
+bool Robot::getFK(const std::vector<double> & joint_point,
```
shaun-edwards: The trailing `const` on a public method indicates it is a thread-safe call.  However, that's not true as other methods may be accessing/changing the robot state variable at the same time.  

For joint solutions (IK), we make the `get` method private and wrap it in a public method called `getJointSolution`.  That method [acquires](https://github.com/shaun-edwards/moveit_simple/pull/19/files#diff-8ae580aedd7e6266898c159ae2e4a380R185) a lock making it thread safe.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -144,6 +144,15 @@ class Robot
                         std::vector<double> & joint_point) const;
 
   /**
+   * @brief getFK finds cartesian pose for the given joint positions.
+   * @param joint_point - joint positions
+   * @param pose - pose corresponding to joint_pose
+   * @return true if pose is found
+   */
+  bool getFK(const std::vector<double> & joint_point,
```
shaun-edwards: Follow the same pattern as `getIK` (see comments below for rationale).  `getIK` is a private method wrapped in a public method called [getJointSolution](https://github.com/shaun-edwards/moveit_simple/pull/19/files#diff-545359f474025b75a7e4eb81e5495903R142)

shaun-edwards (changes requested):
Needs changes to be thread-safe.

shaun-edwards:
@nsnitish, please add tests for new method(s), once complete.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -111,23 +112,30 @@ class Robot
...
    * @param point_name - name of point to add
    * @param time - time from start of trajectory to reach point
+   * @param traj_type - Type of Trajectory from last point to this point
+   * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
    * @throws <std::invalid_argument> (point_name is not found)
    * @throws <tf2::TransformException> (transform of TF named point_name fails)
    */
   void addTrajPoint(const std::string & traj_name, const std::string & point_name,
-                    double time);
+                    double time, const std::string & traj_type = "JOINT",
```
shaun-edwards: Please use an [enumeration\(http://en.cppreference.com/w/cpp/language/enum) rather than strings.  Be sure to declare enumeration in accordance with the [ROS style guide](http://wiki.ros.org/CppStyleGuide#Enumerations)

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -111,23 +112,30 @@ class Robot
...
+                    const unsigned int num_steps = 0);
```
shaun-edwards: Document the behavior of the default value.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -111,23 +112,30 @@ class Robot
    * @param traj_name - name of trajectory buffer to add point to
    * @param point_name - name of point to add
    * @param time - time from start of trajectory to reach point
+   * @param traj_type - Type of Trajectory from last point to this point
```
shaun-edwards: To avoid confusion we should talk about interpolation method.  A trajectory should be reserved for a set of way-points.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -144,12 +152,12 @@ class Robot
                         std::vector<double> & joint_point) const;
 
   /**
-   * @brief getFK finds cartesian pose for the given joint positions.
+   * @brief getPosition finds cartesian pose for the given joint positions.
```
shaun-edwards: I prefer `getPose`, it implies position + orientation.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -88,6 +89,8 @@ void Robot::addTrajPoint(const std::string & traj_name, const Eigen::Affine3d po
     std::unique_ptr<TrajectoryPoint> point =
       std::unique_ptr<TrajectoryPoint>(new CartTrajectoryPoint(pose_rel_robot, time, point_name));
     traj_map_[traj_name].push_back(std::move(point));
+    traj_info_map_[traj_name].first.push_back(std::move(traj_type));
```
shaun-edwards: Are you sure you need to use `std::move`?  This should only be required for `unique_pointers`.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -55,17 +55,16 @@ TEST(MoveitSimpleTest, add_trajectory)
...
   EXPECT_THROW(robot.addTrajPoint("bad_traj", "unknown_name", 1.0),std::invalid_argument);
   EXPECT_THROW(robot.addTrajPoint(TRAJECTORY_NAME, pose, "random_link", 5.0, "unknown_name"), tf2::TransformException);
 
-  ros::Duration(2.0).sleep();  //wait for tf tree to populate
   ROS_INFO_STREAM("Testing trajectory adding of points");
   EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5));
-  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5));
   EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "waypoint1", 1.0));
-  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "tf_pub1",   2.0));
+  EXPECT_NO_THROW(robot.addTrajPoint(TRAJECTORY_NAME, "tf_pub1",   2.0, "CARTESIAN", 5));
```
shaun-edwards: We will need to add a more substantial tests for these addtions.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -50,7 +50,8 @@ class TrajectoryPoint;
 class JointTrajectoryPoint;
 class CartTrajectoryPoint;
 class Robot;
-typedef std::vector<std::unique_ptr<TrajectoryPoint> > Trajectory;
+typedef std::vector<std::unique_ptr<TrajectoryPoint>> Trajectory;
+typedef std::pair<std::vector<std::string>, std::vector<unsigned int>> Trajectory_info;
```
shaun-edwards: Type name should be CamelCase

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -204,6 +272,7 @@ class Robot
 
   // Robot internal objects
   std::map<std::string, Trajectory> traj_map_;
+  std::map<std::string, Trajectory_info> traj_info_map_;
```
shaun-edwards: In just about every place `traj_map_` is referenced, `traj_info_map_` is also used.  This indicates that one or the other is not needed OR they should be combined.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -478,6 +595,23 @@ Eigen::Affine3d Robot::transformToBase(const Eigen::Affine3d &in,
 
 
 
+bool Robot::getFK(const std::vector<double> & joint_point,
+                  Eigen::Affine3d &pose) const
+{
+  robot_state_->setJointGroupPositions(joint_group_, joint_point);
+  const std::vector<std::string> link_names = robot_model_ptr_->getLinkModelNames();
+  if (!link_names.empty())
```
shaun-edwards: You should check that the number of joints is equal to the number of free joints.

shaun-edwards (commented):
I can't quite see how this change allows for mixed, cartesian and joint motion between waypoints.  It appears as though a trajectory can only be one or the other.  Is this true?

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: They can and will be combined in final form. It's just for the review

moveit_simple/src/moveit_simple.cpp
nsnitish: @shaun-edwards ! Is it fine to include a fixed joint in the chain as done in the [kuka resource file](https://github.com/shaun-edwards/moveit_simple/blob/indigo-devel/moveit_simple/test/resources/kuka_kr210/kuka_kr210.srdf#L13). Because if it is then the condition you have mentioned will fail. In this joint model, there is one extra joint ([link_6-tool0](https://github.com/shaun-edwards/moveit_simple/blob/indigo-devel/moveit_simple/test/resources/kuka_kr210/kr210l150.urdf#L95)) apart from 6 active joints

shaun-edwards: The number of free/variable joints should capture this: http://docs.ros.org/indigo/api/moveit_core/html/classmoveit_1_1core_1_1RobotState.html#a711eb6a22c6cc8fbe6de0df2290c5bf0

Any fixed joints should not be included in this count

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -56,6 +56,19 @@ class ExecutionFailureException;
 class IKFailException;
 
 
+enum InterpolationType
```
shaun-edwards: Doesn't follow ROS enum standard...see previous comments for links.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -56,6 +56,19 @@ class ExecutionFailureException;
...
+   {
+      UNKNOWN = 0,
```
shaun-edwards: Good practice!  Creating an unknown value is a great way to catch bugs.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -127,14 +142,16 @@ class Robot
    * @param pose - pose of point to add
    * @param frame - frame (must be a TF accessible frame) in which pose is defined
    * @param time - time from start of trajectory to reach point
-   * @param traj_type - Type of Trajectory from last point to this point
+   * @param type - Type of interpolation from last point to this point
    * By deafult, it is set to JOINT. Can be set to "CARTESIAN" for cartesian Interpolation
+   * @param num_steps - number of points to be interpolated
+   * By deafult, it is set to 0
```
shaun-edwards: But what happens when `num_steps` is set to zero?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -358,62 +353,72 @@ void Robot::execute(const std::string traj_name)
...
   robot_state_->copyJointGroupPositions(joint_group_->getName(), current_joint_position);
   points.push_back(toJointTrajPtMsg(current_joint_position, 0.0));
 
-  for(size_t i = 0; i < traj.size(); ++i)
+  for(size_t i = 0; i < traj_info.size(); ++i)
   {
-    const std::unique_ptr<TrajectoryPoint> & traj_point = traj[i];
-    if (traj_type[i] != "CARTESIAN")
+    const std::unique_ptr<TrajectoryPoint> & traj_point = traj_info[i].point;
+    if (traj_info[i].type == JOINT)
```
shaun-edwards: It's better practice to use a `switch` statement with ENUMs.  Not critical, fix if you want to.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -358,62 +353,72 @@ void Robot::execute(const std::string traj_name)
...
+      if (cartesianInterpolation(traj_point, points, traj_info[i].num_steps))
       {
         ROS_INFO_STREAM("Trajectory successfully added till " << traj_point->name());
       }else{
         ROS_ERROR_STREAM("Conversion to joint trajectory failed for " << traj_name <<
                         " before adding " << traj_point->name());
         return false;
       }
+    }else{
+      ROS_ERROR_STREAM("Unknown interpolation call " << traj_info[i].type);
```
shaun-edwards: Good practice, checking for unknown types.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -443,66 +448,139 @@ bool Robot::jointInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_poi
...
+    std::unique_ptr<TrajectoryPoint> new_point =std::unique_ptr<TrajectoryPoint>
+               ( new CartTrajectoryPoint(new_point_cart->pose(), new_point_cart->time(), ""));
+
+    // Add new point at the end of Joint Trajectory (named points)
+    if(jointInterpolation(new_point, points,(unsigned int) 0))
+    {
+      points_added++;
+      ROS_INFO_STREAM( points_added << " points among " << (num_steps+1) << " added successfullyfor " << traj_point->name());
+    }else{
+      ROS_WARN_STREAM("Conversion to joint trajectory failed at " << points_added << " among "
```
shaun-edwards: Under what conditions can joint interpolation fail?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -443,66 +448,139 @@ bool Robot::jointInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_poi
...
+  {
+    double t = (double)i / (double)(num_steps+1);
+    std::unique_ptr<CartTrajectoryPoint> new_point_cart;
+    interpolate(prev_traj_point,target_point,t,new_point_cart);
+
+    std::unique_ptr<TrajectoryPoint> new_point =std::unique_ptr<TrajectoryPoint>
+               ( new CartTrajectoryPoint(new_point_cart->pose(), new_point_cart->time(), ""));
+
+    // Add new point at the end of Joint Trajectory (named points)
+    if(jointInterpolation(new_point, points,(unsigned int) 0))
```
shaun-edwards: Does this method perform joint interpolation or cartesian interpolation?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -599,8 +677,10 @@ bool Robot::getFK(const std::vector<double> & joint_point,
                   Eigen::Affine3d &pose) const
 {
   robot_state_->setJointGroupPositions(joint_group_, joint_point);
-  const std::vector<std::string> link_names = robot_model_ptr_->getLinkModelNames();
-  if (!link_names.empty())
+  const std::vector<std::string> link_names = joint_group_->getLinkModelNames();
+  const std::vector<std::string> active_joints = joint_group_->getActiveJointModelNames();
+  const int vc =  (int)robot_state_->getVariableCount();
+  if ( active_joints.size() == vc)
```
shaun-edwards: This should always be true.  However, you don't check the `joint_point` parameter, which could be inconsistent.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
-  EXPECT_THROW(robot.execute("traj3"), moveit_simple::IKFailException);
-
   EXPECT_THROW(robot.execute("bad_traj"), std::invalid_argument);
 
 }
 
 
+TEST(MoveitSimpleTest, interpolation)
+{
+  class Robot2: public moveit_simple::Robot
```
shaun-edwards: Use a better name, such at `RobotTest`.  Also, add a comment that indicates why you created this class.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
+  {
+  public:
+    using moveit_simple::Robot::Robot;
```
shaun-edwards: Neat trick!

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
 
-  EXPECT_TRUE(robot.getPosition(joint_point, pose1));
-  EXPECT_TRUE(robot.getJointSolution(pose1, 3.0, seed, joint_point));
-  EXPECT_FALSE(robot.getJointSolution(pose, 3.0, seed, joint_point));
+  EXPECT_TRUE(robot.getPose(joint_point1, pose1));
+  EXPECT_TRUE(robot.getJointSolution(pose1, 3.0, seed, joint_point2));
+  EXPECT_FALSE(robot.getJointSolution(pose, 4.0, seed, joint_point3));
+  EXPECT_TRUE(robot.getPose(joint_point2, pose2));
+
+  double error_pose = (pose1.translation() - pose2.translation()).norm() +
```
shaun-edwards: Are you just trying to compae to Eigen types?  Can you use [isApprox](https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#abf727b0aba068172f17c1725cf5d6819) instead?

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
+                      (pose1.linear() - pose2.linear()).norm();
+  EXPECT_NEAR(error_pose, 0.0, 1e-2);
```
shaun-edwards: If this test fails, it will be nearly impossible to know why.  Can you print out the two eigen transforms (the eigen matrix support the iostream operator, so this shouldn't be too hard).  You can print it out to ROS_DEBUG if it creates too much noise.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
+
+
+  double error_joint = 0.0;
+  for (std::size_t i = 0; i < joint_point1.size(); ++i)
+  {
+    error_joint += fabs(joint_point1[i] - joint_point2[i]);
```
shaun-edwards: Print out the vectors (see comment above).

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -231,16 +259,28 @@ class Robot
...
+
+  /**
+   * @brief interpolate - Joint interpolation from \e from point towards \e to
+   * point at time t in [0,1] and stores it in \e point.
+   * @param from: initial Joint point
+   * @param to: final Joint point
+   * @param t: parameteric time
+   * @param point: interpolated Joint point
+   */
+  void interpolate( const std::unique_ptr<JointTrajectoryPoint>& from,
```
shaun-edwards, line 284: This method doesn't change the underlying class.  It should have the `const` keyword appended to indicate this.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -443,66 +448,139 @@ bool Robot::jointInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_poi
...
+      points.push_back(toJointTrajPtMsg(*new_point));
+      ROS_INFO_STREAM("Appending trajectory point, size: " << points.size());
+      points_added++;
+      ROS_INFO_STREAM( points_added << " points among " << (num_steps+1) <<
+                      " successfully interpolated for " << traj_point->name());
+    }else{
+       ROS_WARN_STREAM("Conversion to joint trajectory failed at " << points_added << " among "
+                        << (num_steps+1) << "points for " << traj_point->name()
+                        << ". Clearing points added in unsuccessful attempt");
+      for (std::size_t j=0; j<points_added; ++j)
```
shaun-edwards: This makes for an awkward method.  Can you think of a better way to write this method that doesn't require you to walk back failures?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -443,66 +448,139 @@ bool Robot::jointInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_poi
...
+    // Add new point at the end of Joint Trajectory (named points)
+    if(jointInterpolation(new_point, points,(unsigned int) 0))
+    {
+      points_added++;
+      ROS_INFO_STREAM( points_added << " points among " << (num_steps+1) << " added successfullyfor " << traj_point->name());
+    }else{
+      ROS_WARN_STREAM("Conversion to joint trajectory failed at " << points_added << " among "
+                        << (num_steps+1) << "points for " << traj_point->name()
+                        << ". Clearing points added in unsuccessful attempt");
+      for (std::size_t j=0; j<points_added; ++j)
```
shaun-edwards: See comment in method above about walking back failures.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
-  EXPECT_NO_THROW(robot.addTrajPoint("traj3", "waypoint2", 1.0));
-  EXPECT_NO_THROW(robot.addTrajPoint("traj3", "waypoint3", 2.0, "CARTESIAN", 1));
-  EXPECT_THROW(robot.execute("traj3"), moveit_simple::IKFailException);
-
   EXPECT_THROW(robot.execute("bad_traj"), std::invalid_argument);
 
 }
 
 
+TEST(MoveitSimpleTest, interpolation)
```
shaun-edwards: This is a good start, but there needs to be more tests here (probably separate tests).  The main missing test is creating a trajectory that is a combination of joint and Cartesian moves and ensuring the whole trajectory is planned correctly.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -50,50 +50,185 @@ TEST(MoveitSimpleTest, reachability)
...
 }
 
+
 TEST(MoveitSimpleTest, add_trajectory)
 {
   moveit_simple::Robot robot(ros::NodeHandle(), "robot_description", "manipulator");
   const std::string TRAJECTORY_NAME("traj1");
   const Eigen::Affine3d pose = Eigen::Affine3d::Identity(); 
   ros::Duration(2.0).sleep();  //wait for tf tree to populate
+  const moveit_simple::InterpolationType cart = moveit_simple::CARTESIAN;
```
shaun-edwards: No need to create temporary variables here.  The enums provide more than enough context to understand what they mean.

shaun-edwards (changes requested):

moveit_simple/test/moveit_simple.cpp
nsnitish: How do we check if it is planned correctly?

shaun-edwards: You can use the same points you use for the interpolate tests.  You might need to rework the class to check the "plan".

moveit_simple/src/moveit_simple.cpp
nsnitish: When we pass the num_steps = 0 to jointInterpolation it only appends the new_point to the points

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: It just appends the point

moveit_simple/src/moveit_simple.cpp
shaun-edwards: Do you add the point `num_steps` times?  This section is in the `for` loop above.

moveit_simple/src/moveit_simple.cpp
nsnitish: If the point to be added is cartesian then it can fail for IK call.

moveit_simple/src/moveit_simple.cpp
nsnitish: It adds `num_steps` interpolated points + traj_point from Trajectory (added with addTrajPoint)

moveit_simple/test/moveit_simple.cpp
nsnitish: Does checking the points in the method [toJointTrajectory](https://github.com/shaun-edwards/moveit_simple/pull/19/files#diff-8ae580aedd7e6266898c159ae2e4a380R353) enough to verify that? This is the exact method that is "planning".

moveit_simple/src/moveit_simple.cpp
nsnitish: Need to address this

moveit_simple/src/moveit_simple.cpp
nsnitish: Need to address this

shaun-edwards (approved):

Issue #10, opened 2017-05-05
Add method to get current robot position


shaun-edwards:
It would be convenient to be able to get the current robot position from the `moveit_simple` object.  The current implementation uses the moveit RobotState to both capture the current state and seed IK calls.  This is the root cause of issue #3.  I think we can address that issue and this one at the same time.

Issue #4, opened 2017-05-04
Break up moveit simple into offline robot and online robot object.


shaun-edwards:
The `moveit_simple` object is currently used in two different manors.  The first way uses `moveit_simple` to perform kinematics and path planning.  The second way builds on top of this functionality to execute paths (online).  Specifically the [execute method](https://github.com/shaun-edwards/moveit_simple/blob/indigo-devel/moveit_simple/include/moveit_simple/moveit_simple.h#L150) and the [updateState](https://github.com/shaun-edwards/moveit_simple/blob/indigo-devel/moveit_simple/include/moveit_simple/moveit_simple.h#L198) are used exclusively for online functionality.  The big difference between offline and online functionality is that the current robot state (online) is often used to seed various IK calls.  When used in an offline manor, how the seed is determines is unclear (in reality it simply uses the joint pose calculated from the last call to IK, which might not be expected).

shaun-edwards:
@geoffreychiou, could you look at this one?

nsnitish:
I think we have already addressed this in #21

Issue #3, opened 2017-05-03
Copying current pose to joint trajectory doesn't always work


shaun-edwards:
When executing a joint trajectory, the first pose is copied from the current robot state [here](https://github.com/shaun-edwards/moveit_simple/blob/bfd84cd29ca29ad9128ce31c5a5bf2bd10fe2484/moveit_simple/src/moveit_simple.cpp#L354).  If the robot state has not been updated via the `joint_states` message since the last IK call, then the wrong joint pose is captured.

shaun-edwards:
@geoffreychiou, could you take a look at this one next week?

geoffreychiou:
@shaun-edwards 

I'm trying to write the unit-test for this.

Lets say i'm running this:
``` C++
  robot.addTrajPoint(TRAJECTORY_NAME, "home",      0.5);
  robot.addTrajPoint(TRAJECTORY_NAME, "waypoint1", 1.0);
  robot.addTrajPoint(TRAJECTORY_NAME, "tf_pub1",   2.0);
  robot.addTrajPoint(TRAJECTORY_NAME, "waypoint2", 3.0);
  robot.addTrajPoint(TRAJECTORY_NAME, "waypoint3", 4.0);

  robot.execute(TRAJECTORY_NAME);
  robot.execute(TRAJECTORY_NAME);
```

The expected behavior is:
``` C++
// Robot Joints @ [0 0 0 0 0 0] (home)
  robot.execute(TRAJECTORY_NAME);
  {
    // Robot Joints Start @ [0 0 0 0 0 0] (home)
    // Robot runs through trajectory
    // Robot Joints End @ [waypoint3]
  }
// Robot Joints @ [waypoint3]
  robot.execute(TRAJECTORY_NAME);
  {
    // Robot Joints Start @ [waypoint3]
    // Robot runs through trajectory
    // Robot Joints End @ [waypoint3]
  }
```

Instead, the robot state is not updated all the time so the second execution might start from home instead of waypoint3?

geoffreychiou:
Here is debugging output without any fix:

![](http://i.imgur.com/tHU4l3l.png)

I ran execute 4 times and published the joint values before and after each execution. On the lines where there are two joints in a row, I just published the joint values twice and you will notice that it is the correct value on the second one.

My fix was to add the first two lines to the code below:

``` C++
  ros::spinOnce();
  ros::Duration(0.100).sleep();

[...]

PR #21, opened 2017-08-05
Added getCurrentRobotState method to address Issue #3, #4 and #10

nsnitish:
Separated the current_robot_state_ (stores the current position of the robot) from robot_state_ (used for FK and IK calls)


nsnitish:
@shaun-edwards ! Can you please have a look?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -226,6 +226,15 @@ class Robot
...
                                   double t) const;
 
+
+  /**
+  * @brief getCurrentRobotState - Returns a vector<double> of the
+  * current joint positions of the robot from robot_state_.
+  *
+  * @return std::vector<double> current_joint_position
+  */
+  std::vector<double> getCurrentRobotState(void) const;
```
shaun-edwards: Let's call this `getJointState`.  I know MoveIt calls it `RobotState`, but that's confusing to a more traditional robot programmer.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -315,12 +324,14 @@ class Robot
 
   void updateState(const sensor_msgs::JointStateConstPtr& msg);
 
+  void updateCurrentState(const sensor_msgs::JointStateConstPtr& msg);
 
   // Robot internal objects
   std::map<std::string, TrajectoryInfo> traj_info_map_;
 
   // MoveIt objects
   mutable moveit::core::RobotStatePtr robot_state_;
```
shaun-edwards: I realize this is the name that I chose, but now that there is two "robot states", the names should indicate the difference.  That isn't the case right now.  Can you think of new names?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -770,6 +770,21 @@ void Robot::updateState(const sensor_msgs::JointStateConstPtr& msg)
...
+  std::lock_guard<std::recursive_mutex> guard(m_);
+  ros::spinOnce();
+  std::vector<double> current_joint_positions;
+  current_robot_state_->update();
+  current_robot_state_->copyJointGroupPositions(joint_group_->getName(), current_joint_positions);
+  return current_joint_positions;
+}
+
+
+void Robot::updateCurrentState(const sensor_msgs::JointStateConstPtr& msg)
```
shaun-edwards: Why not stick to a `vector<double>`?  I prefer to not pollute this class with ROS message types except where necessary.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -24,6 +24,23 @@ using testing::Types;
 namespace moveit_simple_test
 {
 
+
+std::string vector_to_string(const std::vector<double> &input_vector)
```
shaun-edwards: This should not be needed.  The pretty print header takes care of this.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -289,10 +306,10 @@ TEST(MoveitSimpleTest, interpolation)
   for (std::size_t i = 0; i < joint_expected.size(); ++i)
   {
     error_joint += fabs(joint_out[i] - joint_expected[i]);
-    ROS_INFO_STREAM(" joint_out_ " << i << joint_out[i]);
-    ROS_INFO_STREAM(" joint_expected_ " << i << joint_expected[i]);
   }
   EXPECT_NEAR(error_joint, 0.0, 1e-2);
+  ROS_INFO_STREAM("joint_out" << vector_to_string(joint_out));
```
shaun-edwards: Have you tried using just the stream operator, `<<`?  The prettyprint header should automatically make this work.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -419,4 +435,49 @@ TEST(MoveitSimpleTest, kinematics)
 
 }
 
+
+TEST(MoveitSimpleTest, copy_current_pose)
+{
+  moveit_simple::Robot robot(ros::NodeHandle(), "robot_description", "manipulator");
```
shaun-edwards: This test and others has a lot of the same setup.  Look into Gtest Fixtures to eliminate some of this common code.

shaun-edwards (commented):
Couple of issues to address.  Also, the travis CI is failing.

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: One says robot_state_ and other one is called current_robot_state_. Should we name them like online_robot_state_ and virtual_robot_state_?

moveit_simple/src/moveit_simple.cpp
nsnitish: @shaun-edwards! You are referring to not using the ROS messages in [updateState](https://github.com/shaun-edwards/moveit_simple/pull/21/files#diff-8ae580aedd7e6266898c159ae2e4a380R767), right?

nsnitish:
@shaun-edwards! Not sure why travis CI didn't run test this time!

moveit_simple/src/moveit_simple.cpp
shaun-edwards: yep.

shaun-edwards:
@nsnitish, please resolve conflicts

nsnitish:
Please don't merge this PR as of now. Still need to update few things.

moveit_simple/src/moveit_simple.cpp
nsnitish: Just realized, we just need only current state to be updated through the NodeHandle subscribe.

nsnitish:
@shaun-edwards! Please have a look.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -353,15 +341,68 @@ class Robot
...
+
+protected:
+  OnlineRobot();
+
+  void updateCurrentState(const sensor_msgs::JointStateConstPtr& msg);
+  mutable moveit::core::RobotStatePtr current_robot_state_;
+
+
+  // Visualizations
+  moveit_visual_tools::MoveItVisualToolsPtr visual_tools_;
```
shaun-edwards: Should visual tools be in the `Robot` class?  If I remember correctly, this object provides a visual or the robot state.  Maybe we should have one for the 'real' robot and one for the 'virtual' robot.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -799,14 +803,34 @@ trajectory_msgs::JointTrajectoryPoint Robot::toJointTrajPtMsg(
...
+  std::vector<double> current_joint_positions;
+  virtual_robot_state_->update();
+  virtual_robot_state_->copyJointGroupPositions(joint_group_->getName(), current_joint_positions);
+  return current_joint_positions;
+}
+
+
 
-void Robot::updateState(const sensor_msgs::JointStateConstPtr& msg)
+std::vector<double> OnlineRobot::getJointState(void) const
```
shaun-edwards, line 872: Since this method overries the `Robot::getJointState` do we lose access to the virtual robot joint position?  Is this what you intended?

moveit_simple/test/moveit_simple.cpp
```diff
@@ -25,23 +25,78 @@ using testing::Types;
...
 TEST(MoveitSimpleTest, reachability)
 {
-  moveit_simple::Robot robot(ros::NodeHandle(), "robot_description", "manipulator");
-  const Eigen::Affine3d pose = Eigen::Affine3d::Identity(); 
-  ros::Duration(2.0).sleep();  //wait for tf tree to populate
+  moveit_simple::Robot robot("robot_description", "manipulator");
 
   ROS_INFO_STREAM("Testing reachability of unknown point, should fail");
   EXPECT_FALSE(robot.isReachable("unknown_name"));
-  EXPECT_FALSE(robot.isReachable(pose, "random_link"));
```
shaun-edwards: Why were these deleted?

moveit_simple/test/moveit_simple.cpp
```diff
@@ -25,23 +25,78 @@ using testing::Types;
 namespace moveit_simple_test
 {
 
+/**
+ * @brief UserRobotTest is a fixture for testing public methods.
+ * Objects that can be directly used inside the test
+ - robot pointer for moveit_simple::OnlineRobot
+*/
+class UserRobotTest : public ::testing::Test
```
shaun-edwards, line 33: +1 for test fixtures!

shaun-edwards (commented):
Pretty good!...just a few comments/questions.

moveit_simple/src/moveit_simple.cpp
nsnitish: virtual robot joint position will be there but that will be used for running the IK and FK. When we have online robot, seed conditions are taken from online robot joint position but otherwise virtual robot joint positions should be used.

moveit_simple/test/moveit_simple.cpp
nsnitish: My bad. I'll re-add them.

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: visual_tools_ uses `NodeHandle` to publish but `Robot` class doesn't have any `NodeHandle`. Isn't this required? Why do we need one for 'virtual' robot?

shaun-edwards: It visualize the robot state, which `Robot` has.  It's useful for debugging.  Although it's not explicit, the MoveIt `RobotState` already requires a `NodeHandle`, it's just hidden.

moveit_simple/src/moveit_simple.cpp
shaun-edwards: Is this a publc method?  The inheritance rules on this are confusing for most developers...I recommend a clear separation so people know what they are calling.

nsnitish: I think making the getJointState a protected method in Robot and public in OnlineRobot makes more sense. Because JointState in Robot doesn't make much sense for user.

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: Do we also need separate NodeHandles for online and offline robot?

shaun-edwards: Node handles are singletons...no matter how many you create, there is only ever one.  It's safe to the use the same one for both.

moveit_simple/src/moveit_simple.cpp
shaun-edwards: +1.

nsnitish:
@shaun-edwards! Please have a look

shaun-edwards (approved):

Issue #9, opened 2017-05-05
execute method does not check trajectory points for collision


shaun-edwards:
The execute method does not check the trajectory for collisions.  The transformation into joint points is done [here](https://github.com/shaun-edwards/moveit_simple/blob/bfd84cd29ca29ad9128ce31c5a5bf2bd10fe2484/moveit_simple/src/moveit_simple.cpp#L346-L417).  Collision checking can be easily done similar to [here](https://github.com/shaun-edwards/moveit_simple/blob/bfd84cd29ca29ad9128ce31c5a5bf2bd10fe2484/moveit_simple/src/moveit_simple.cpp#L218-L219).  I would suggest we pass a `collision_check` boolean to the `execute` method that turns collision checking on/off.

PR #22, opened 2017-08-09
Collision check in the execute method (Issue #9)

nsnitish:
1. Added meshes for collision check and visual
2. collision check before adding the interpolated points to [points](https://github.com/shaun-edwards/moveit_simple/blob/bfd84cd29ca29ad9128ce31c5a5bf2bd10fe2484/moveit_simple/src/moveit_simple.cpp#L347).


moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -187,7 +195,7 @@ class Robot
    * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
    * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
    */
-  void execute(const std::string traj_name);
+  void execute(const std::string traj_name, bool collision_check = false);
```
shaun-edwards: Update comments.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -85,6 +85,14 @@ class Robot
...
   Robot(const ros::NodeHandle & nh, const std::string &robot_description,
         const std::string &group_name);
+  /**
+   * @brief isInCollision  returns true if joint_point results in robot config that is
+   * in collision with the environment as defined by the URDF.
+   * @param joint_point(optional) - joint position of the robot to check
+   * If no joint point is provided current position is checked for collision
+   * @return
+   */
+  bool isInCollision(std::vector<double> joint_point = std::vector<double>() ) const;
```
shaun-edwards: Pass by reference.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -212,6 +212,25 @@ bool Robot::getPose(const std::vector<double> & joint_point,
...
+
+bool Robot::isInCollision(std::vector<double> joint_point) const
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+
+  std::vector<double> local_joint_point = joint_point;
+
+  if (joint_point.empty())
+  {
+    ROS_INFO_STREAM("Empty joint point passed to isIncollision, using current state");
```
shaun-edwards: This method can get called a lot.  I would make this a debug message.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -328,27 +347,34 @@ void Robot::execute(const std::string traj_name)
...
+        throw IKFailException("Conversion to joint trajectory failed for " + traj_name);
       }
     }
-    else
+    catch(CollisionDetected &cd)
     {
-      ROS_ERROR_STREAM("Failed to convert " << traj_name << " to joint trajectory");
-      throw IKFailException("Conversion to joint trajectory failed for " + traj_name);
+      ROS_ERROR_STREAM("Collision detected in the " << traj_name);
+      throw cd;
```
shaun-edwards, line 376: Is there a good reason to catch this and re-throw.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -246,7 +255,7 @@ class Robot
    */
   bool jointInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_point,
            std::vector<trajectory_msgs::JointTrajectoryPoint> & points,
-           const unsigned int num_steps);
+           const unsigned int num_steps, bool collision_check = false);
```
shaun-edwards, line 262: Are you sure you want to check for collisions here?  Isn't this an internal method?  It would be more straight-forward to do this check once, in the plan/execute method.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -258,7 +267,7 @@ class Robot
    */
   bool cartesianInterpolation(const std::unique_ptr<TrajectoryPoint> & traj_point,
            std::vector<trajectory_msgs::JointTrajectoryPoint> & points,
-           const unsigned int num_steps);
+           const unsigned int num_steps, bool collision_check = false);
```
shaun-edwards, line 275: See comment above.

shaun-edwards (commented):
Some questions/comments to be addressed.

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: We are storing all the planned jointTrajectoryPoints in "points". So, it should check for collision before adding them to points.

moveit_simple/src/moveit_simple.cpp
nsnitish: It's actually only for logging the traj_name in which the collision has occurred.

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: Do you have to add a collision check to multiple methods, or is simpler, cleaner, and easier to maintain if we check in one place.

nsnitish: It is only checked [here once](https://github.com/shaun-edwards/moveit_simple/pull/22/files#diff-8ae580aedd7e6266898c159ae2e4a380R503).

shaun-edwards (approved):

Issue #11, opened 2017-05-05
Add isNearSingular method


shaun-edwards:
A method is needed that returns true when the robot arm configuration is near singular.

nsnitish:
@shaun-edwards! Determinant of the [Jacobian](http://docs.ros.org/jade/api/moveit_core/html/classmoveit_1_1core_1_1RobotState.html#af14fd5a53bf052f12eeb81237dbdc67b) is what we are concerned here, right?

shaun-edwards:
Yes...although @geoffreychiou expressed some reservation about using the jacobian to determine singularity.

nsnitish:
@geoffreychiou Can you please elaborate?

geoffreychiou:
Basically a singularity occurs when the determinant of the Jacobian is equal to zero, but I found a article/stackoverflow  comment a while back saying that you can't just take the determinant of J and check if it is close to zero.

One possibility is to compare the joint velocities to the Cartesian velocity. When the robot is near a singularity, the joint velocities should be very large compared to the Cartesian velocity. It would be up to you to determine what the threshold should be.

PR #23, opened 2017-08-16
Added isNearSingular method (Issue #11)

nsnitish:
Issue #11


nsnitish:
@shaun-edwards! Can you please have a look?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -775,6 +775,44 @@ bool Robot::getIK(const Eigen::Affine3d pose, std::vector<double> & joint_point,
 
 
 
+bool Robot::isNearSingular(const std::vector<double> & joint_point) const
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+  double threshold = 1e-10; // How to determine this threshold?
```
shaun-edwards: This could also be a parameter you pass in, but I think the user will be more confused by this than we are : /

moveit_simple/src/moveit_simple.cpp
```diff
@@ -775,6 +775,44 @@ bool Robot::getIK(const Eigen::Affine3d pose, std::vector<double> & joint_point,
...
+  if (robot_state_->getJacobian(joint_group_, joint_group_->getLinkModels().back(),
+                                  reference_point, jacobian, false))
+  {
+    // Check for determinant of J*J'
+    if (fabs((jacobian*jacobian.transpose()).determinant()) <= threshold)
+      {
+        ROS_WARN_STREAM("Given configuration " <<local_joint_point <<
+                                                        " is near Singularity");
+        return true;
+      }else{
```
shaun-edwards: Correct formatting, here and elsewhere.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -775,6 +775,44 @@ bool Robot::getIK(const Eigen::Affine3d pose, std::vector<double> & joint_point,
...
+        ROS_INFO_STREAM("Given configuration " <<local_joint_point <<
+                                               " is away from any Singularity");
+        return false;
+      }
+  }else{
+    ROS_ERROR_STREAM("Jacobian not found for " << local_joint_point);
+    throw std::invalid_argument("joint_group " + joint_group_->getName() +
```
shaun-edwards: Can this ever happen?

Since the "joint group" is never passed in to this method, throwing an invalid argument exception doesn't make much sense.

moveit_simple/test/moveit_simple.cpp
```diff
@@ -468,4 +468,23 @@ TEST(MoveitSimpleTest, collision)
...
+TEST(MoveitSimpleTest, Singularity)
+{
+  moveit_simple::Robot robot(ros::NodeHandle(), "robot_description", "manipulator");
+  ros::Duration(2.0).sleep();  //wait for tf tree to populate
+
+  std::vector<trajectory_msgs::JointTrajectoryPoint> points;
+
+  std::vector<double>joint1(6,0);
+  std::vector<double>joint2(6,0);
+  joint1[4] = M_PI/2;
```
shaun-edwards: Add a comment that explains why these are or are not singular?

shaun-edwards: There are a fixed number of singular configurations for standard 6DOF industrial arms.  Perhaps you should test them all.

shaun-edwards (commented):
Some questions and requests.

moveit_simple/test/moveit_simple.cpp
nsnitish: Do you have any good reference for singularities of KUKA kr210?

moveit_simple/src/moveit_simple.cpp
nsnitish: Ok. Then probably I should get rid of this. But are you sure moveit_simple won't be extended to cases where "joint group" won't be passed as a chain?

moveit_simple/src/moveit_simple.cpp
nsnitish: I am not able to find any reference on this which talks about this threshold. What is the expected angle difference from the singular configuration that you think we should target to trigger isNearSingular? That can help in deciding a safe threshold.

moveit_simple/test/moveit_simple.cpp
shaun-edwards: Whenever 4 & 6 align, 6 parallel to the world above axis 1, whenver the "elbow" is straight, any configs at the "extent" of the reach.

moveit_simple/src/moveit_simple.cpp
shaun-edwards: Go ahead and log the error and just return false.  Let's not worry about what may be in the future.

moveit_simple/src/moveit_simple.cpp
shaun-edwards: I would pick a robot config, say axis 4 and 6 within 3 degrees of being aligned and see what the threshold should be set to in order to catch this.

moveit_simple/test/moveit_simple.cpp
nsnitish: So far, I am just testing 4 & 6 alignment. Will add the remaining tests as well

moveit_simple/src/moveit_simple.cpp
nsnitish: Sounds good

nsnitish:
@shaun-edwards ! Please have a look

shaun-edwards (approved):

Issue #20, opened 2017-07-15
Support for custom tool frame


shaun-edwards:
The current code assumes the tool point is the same as the moveit end-link.  This makes reusing IK solvers difficult.  Instead the system should accept a custom tool frame name and perform the necessary frame conversions itself.

shaun-edwards:
@MauleshSTrivedi, can you please look at this issue?

MauleshSTrivedi:
@shaun-edwards I will get onto this right away.

shaun-edwards:
FYI @MauleshSTrivedi, I was able to run the `moveit_simple` tests using `catkin run_tests`.  I had some trouble with `catkin tools`, but they seem to work on Travis.

BTW, I forgot to mention, the following is also needed to build/test `moveit_simple`: https://github.com/shaun-edwards/prettyprint

MauleshSTrivedi:
@shaun-edwards Yeah I cloned prettyprint so that's fine. The test are running for me, but I am getting the overall results as a FAILURE.. Basically, 6/9 tests PASSED while 3 of them generated a FAILURE. I will look at your latest build results to corroborate.

There are some other errors that I seem to be getting (ROS Errors and not test errors), and I need to look into that as well.

shaun-edwards:
Some ROS errors are a normal part of the tests. I have a half dozen or so.

MauleshSTrivedi:
Got it. Will take that into consideration.

shaun-edwards:
FYI...here's my output

```
shaun@shaun-Precision-5510:~/plusone/moveit_simple$ catkin_make run_tests
Base path: /home/shaun/plusone/moveit_simple
Source space: /home/shaun/plusone/moveit_simple/src
Build space: /home/shaun/plusone/moveit_simple/build
Devel space: /home/shaun/plusone/moveit_simple/devel
Install space: /home/shaun/plusone/moveit_simple/install
####
#### Running command: "make cmake_check_build_system" in "/home/shaun/plusone/moveit_simple/build"
####
####
#### Running command: "make run_tests -j8 -l8" in "/home/shaun/plusone/moveit_simple/build"
####
Scanning dependencies of target clean_test_results_moveit_simple

<build steps>

Scanning dependencies of target _run_tests_moveit_simple_rostest_test_launch_utest.launch
-- run_tests.py: execute commands
[...]

PR #26, opened 2017-10-10
Added customToolFrameTF() function to address Issue #20.

MauleshSTrivedi:
Added support for custom tool frame to solve IK.


nsnitish:
@MauleshSTrivedi ! I think you have uploaded some cmake and catkin generated files which should not be here. Please get rid of those. Also, the travis is failing.

MauleshSTrivedi:
@nsnitish Oh, my bad, I didn't realize. I'll get on it in a bit. I will update this PR by tomorrow morning.

shaun-edwards:
@MauleshSTrivedi, always review the PR changes before you submit.  Mistakes such as this are obvious in the changes tab.

gavanderhoorn:
@MauleshSTrivedi: out of curiosity: how did you setup your workspace? Seeing those CMake configuration phase files in your PR makes me wonder if you are using the `moveit_simple` package root as the root of your workspace.

How are you building things?

MauleshSTrivedi:
@gavanderhoorn I think these files crept in when I tried to sync my project with QT (as far as I can remember).. my project setup looks something like this -- /home/mauleshtrivedi/foo_projects/moveit_simple_workspace/src/moveit_simple
/home/mauleshtrivedi/foo_projects/moveit_simple_workspace/src/prettyprint

And the only git repositories in my foo_projects are moveit_simple and prettyprint. Does that help?

gavanderhoorn:
qtc will do that to you, yes.

Have you looked at @levi-armstrong's plugin?

MauleshSTrivedi:
@gavanderhoorn I have not  looked into it but that's a good idea.. let me try that later with another test project.. for now, I need to find a way to remove these autogenerated files.. I'm thinking of just creating a fresh repo in the worst case.. shouldn't take time.

gavanderhoorn:
Commit / stash any outstanding changes, make sure you don't have any untracked files that should really be tracked, then run a `git clean -fnd` and inspect the output.

If it's ok, run `git clean -fd` and your clones should be back to their `HEAD`.

MauleshSTrivedi:
@gavanderhoorn git clean returns:
Would remove moveit_simple/CMakeFiles/CMakeTmp/
Would remove moveit_simple/gtest/
Would remove moveit_simple/test_results/

But there are more than these 3 directories that QT created.. I'll remove these from my commit for now, but I still think it would be a better idea to clone moveit_simple from scratch just to be sure (I might be overthinking)

gavanderhoorn:
Up to you.

MauleshSTrivedi:
@shaun-edwards So sorry for the mess.. I should've been more careful before submitting the PR. I'll make sure to double check travis before submitting a request.

MauleshSTrivedi:
@gavanderhoorn Thank you so much. You have been most helpful. I've installed Levi's ppa and I'll test it with qt in a bit.

shaun-edwards:
+1 to using Levi's QT plugin

PR #27, opened 2017-10-10
Added customToolFrameTF() function to address Issue #20.

MauleshSTrivedi:
TODO: Add Unit tests.

Added functionality for setting a goal_pose from a custom tool frame in getJointSolution(...). The function customToolFrameTF(...) takes a target_pose and a tool_frame as input arguments and computes the TF from this frame to the moveit_end_link. The transformed goal_pose is then used for calculating the IK. 

The next step would be to add this functionality to the isReachable(...) and isInCollision(...) methods.


MauleshSTrivedi:
testing some discrepancies. sending up a new PR that only slightly modifies the above one.

PR #28, opened 2017-10-11
Added transformToolToEndLink() function to address Issue #20

MauleshSTrivedi:
TODO: Add Unit tests.

Added functionality for setting a goal_pose from a custom tool frame in getJointSolution(...). The function transformToolToEndLink(...) takes a target_pose and a tool_frame as input arguments and computes the TF from this frame to the moveit_end_link. The transformed goal_pose is then used for calculating the IK.


moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -323,6 +344,7 @@ class Robot
   bool getFK(const std::vector<double> & joint_point,
                                Eigen::Affine3d &pose) const;
 
+  /*Lookup Trajectory Point in the Robot Model*/
```
shaun-edwards: Document this method or remove the comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -152,6 +152,7 @@ void Robot::addTrajPoint(const std::string & traj_name,
   traj_info_map_[traj_name].push_back({std::move(point), type, num_steps});
 }
 
+/*Lookup Trajectory Point in the Robot Model*/
```
shaun-edwards: Remove the comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
   return getIK(pose, local_seed, joint_point, timeout);
 }
 
+/*------------------------------------------------------------------------------------------------------------------------------*/
+
+/*ISSUE #20 -- getJointSolution() WITH CUSTOM TOOL FRAME FOR SOLVING IK*/
```
shaun-edwards: Remove comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  std::vector<double> local_seed = seed;
+  if (seed.empty())
+  {
+    ROS_INFO_STREAM("Empty seed passed to getJointSolution, using current state");
+    local_seed =  getJointState();
+  }
+  return getIK(custom_frame_goal_pose, local_seed, joint_point, timeout);
+}
+
+/* Returns the goal pose transformed to moveit_end_link from Custom tool frame*/
```
shaun-edwards: Remove comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+Eigen::Affine3d Robot::transformToolToEndLink(const Eigen::Affine3d &target_pose,
+                                              const std::string& custom_tool_frame) const
+{
+  
+  if(custom_tool_frame.compare(joint_group_->getEndEffectorName()) == 0 &&
+     robot_model_ptr_->hasLinkModel(custom_tool_frame)) {
```
shaun-edwards: The ROS style is to put the '{' on the following line.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+
+      ROS_INFO_STREAM("Custom Tool Frame is same as moveit end_link");
```
shaun-edwards: The log message doesn't indicate what is happening...you should add something like "returning same target pose"

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  if (seed.empty())
+  {
+    ROS_INFO_STREAM("Empty seed passed to getJointSolution, using current state");
+    local_seed =  getJointState();
+  }
+  return getIK(custom_frame_goal_pose, local_seed, joint_point, timeout);
+}
+
+/* Returns the goal pose transformed to moveit_end_link from Custom tool frame*/
+Eigen::Affine3d Robot::transformToolToEndLink(const Eigen::Affine3d &target_pose,
```
shaun-edwards: If it doesn't lead to too many branches, I always recommend returning from a single point in a function.  This function returns from two different places.  This makes it hard to debug.  Say you wanted to log the return of this method, you would have to track down all the `return` statements and add log messages to each.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  
+  if(custom_tool_frame.compare(joint_group_->getEndEffectorName()) == 0 &&
+     robot_model_ptr_->hasLinkModel(custom_tool_frame)) {
+
+      ROS_INFO_STREAM("Custom Tool Frame is same as moveit end_link");
+
+      return target_pose;
+    }
+
+  if( tf_buffer_.canTransform(joint_group_->getEndEffectorName(), custom_tool_frame, ros::Time::now(), ros::Duration(0.1)) ) {
```
shaun-edwards: What happens if this isn't true?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+    try
+    {
+        ROS_INFO_STREAM("Looked up tf named frame: " << custom_tool_frame);
+
+        geometry_msgs::TransformStamped trans_msg;
+        Eigen::Affine3d pose_buffer = target_pose;
+        trans_msg = tf_buffer_.lookupTransform(joint_group_->getEndEffectorName(), 
+                                               custom_tool_frame, ros::Time::now(),
```
shaun-edwards: No need to wait for `ros::Time::now()` since we expect all these transforms to be static.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  if( tf_buffer_.canTransform(joint_group_->getEndEffectorName(), custom_tool_frame, ros::Time::now(), ros::Duration(0.1)) ) {
+    try
+    {
+        ROS_INFO_STREAM("Looked up tf named frame: " << custom_tool_frame);
+
+        geometry_msgs::TransformStamped trans_msg;
+        Eigen::Affine3d pose_buffer = target_pose;
+        trans_msg = tf_buffer_.lookupTransform(joint_group_->getEndEffectorName(), 
+                                               custom_tool_frame, ros::Time::now(), 
+                                               ros::Duration(5.0));
```
shaun-edwards: What happens if the timeout expires?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,9 +218,70 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+          
+        tf::transformMsgToEigen(trans_msg.transform, pose_buffer);
+        ROS_INFO_STREAM("Using TF to lookup transform " << custom_tool_frame << " frame: " << std::endl << pose_buffer.matrix());
+          
+        return pose_buffer;
+    }
+    catch (tf2::TransformException &ex)
+    {
+        ROS_ERROR_STREAM("TF transform lookup failed: " << ex.what());
+        throw ex;
```
shaun-edwards: Should you `throw` from this method?  The method returns a boolean.  Why not use that.

shaun-edwards (changes requested):

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Ok, yeah that makes sense.. but in that case, would you want me to modify some of the other functions that are designed in a similar way? (Ex.: lookupTrajectoryPoint(...)

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Ok, I'll keep that in mind :)

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: I kept this condition here as a base-check for now.. I'm still working on whether this is needed OR lookUpTransform() will throw out the same exceptions if we should fail. I'll modify this accordingly.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Ok so if lookup fails, I'm throwing an exception on the error-stream. Calling it a fail for now.. Would you rather I change this design?

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: I'm not sure if there is a function call that does not need time, but I can look into it.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: The method getJointSolution(...) returns a boolean. We expect this method to return an Eigen::Affine3d

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: Ok

shaun-edwards:
@MauleshSTrivedi, is this ready to be reviewed?

MauleshSTrivedi:
@shaun-edwards Yeah it is. The unit tests are also running successfully and I have made the necessary changes you requested in the last review (minus a couple where I wasn't sure, but I left comments for you).

MauleshSTrivedi:
@shaun-edwards Thanks!

moveit_simple/CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ find_package(catkin REQUIRED COMPONENTS
 )
 
 find_package(Boost REQUIRED)
-find_package(Eigen REQUIRED)
+find_package(Eigen3 REQUIRED)
+# find_package(Eigen REQUIRED) --> FindEigen.cmake Module has been depricated in Kinetic, please use FindEigen3.cmake Module [find_package(Eigen3)]
```
shaun-edwards: Did you leave this comment in for some reason?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
   return getIK(pose, local_seed, joint_point, timeout);
 }
 
+bool Robot::getJointSolution(const Eigen::Affine3d &pose, const std::string& custom_tool_frame,
+                             double timeout, const std::vector<double> & seed,
+                             std::vector<double> & joint_point) const
+{
+
+  // Ideally "tool0" should be replaced with joint_group_->getEndEffectorName()
+  std::string moveit_tool_link = "tool0";
```
shaun-edwards: This needs to be fixed before this is accepted.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  if(frame_in.compare(frame_out) == 0 && 
+    robot_model_ptr_->hasLinkModel(frame_in) &&
+    robot_model_ptr_->hasLinkModel(frame_out)) {
+
+      ROS_INFO_STREAM("Returning same target_pose");
+
+      return target_pose;
+  }
+
+  //if( tf_buffer_.canTransform(joint_group_->getEndEffectorName(), ros::Time::now(), ros::Duration(0.1)) ) {
```
shaun-edwards: Delete this comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+Eigen::Affine3d Robot::transformPoseBetweenFrames(const Eigen::Affine3d &target_pose,
+                                              const std::string& frame_in,
+                                              const std::string& frame_out) const
+{
+
+  if(frame_in.compare(frame_out) == 0 && 
+    robot_model_ptr_->hasLinkModel(frame_in) &&
+    robot_model_ptr_->hasLinkModel(frame_out)) {
+
+      ROS_INFO_STREAM("Returning same target_pose");
```
shaun-edwards: Please expand this log message with some more details...frame_in, frame_out.  This message won't help much in the log (out of the context of the code).

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+
+      return target_pose;
+  }
+
+  //if( tf_buffer_.canTransform(joint_group_->getEndEffectorName(), ros::Time::now(), ros::Duration(0.1)) ) {
+    try
+    {
+        ROS_INFO_STREAM("Looked up tf named frame: " << frame_in);
```
shaun-edwards: Add more context here as well....looking up frame because it doesn't exist in Robot model.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+
+        geometry_msgs::PoseStamped pose_stamped_buffer;
+        geometry_msgs::PoseStamped transformed_pose;
+        geometry_msgs::TransformStamped trans_msg;
+        Eigen::Affine3d target_pose_buffer = target_pose;
+
+        trans_msg = tf_buffer_.lookupTransform(frame_out,
```
shaun-edwards: Please see my previous comment.  There should be a variant of `lookupTransform` that returns the latest version (doesn't take a time argument).

shaun-edwards: Or takes a time of "zero"

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+bool Robot::getJointSolution(const Eigen::Affine3d &pose, const std::string& custom_tool_frame,
+                             double timeout, const std::vector<double> & seed,
+                             std::vector<double> & joint_point) const
+{
+
+  // Ideally "tool0" should be replaced with joint_group_->getEndEffectorName()
+  std::string moveit_tool_link = "tool0";
+
+  // Transform Target/Goal Point from custom tool frame to moveit_end_link
+  const Eigen::Affine3d custom_frame_goal_pose =
```
shaun-edwards: `transformPoseBetweenFrames` throws...should you catch it an handle the boolean return.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+        ROS_INFO_STREAM("Using TF to lookup transform " << frame_out << " frame: " << std::endl << target_pose_buffer.matrix());
+
+        return target_pose_buffer;
+    }
+    catch (tf2::TransformException &ex)
+    {
+        ROS_ERROR_STREAM("TF transform lookup failed: " << ex.what());
+        throw ex;
+    }
+  //}
```
shaun-edwards: Remove comment.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  
+  Eigen::Affine3d pose_buffer;
+
+  // Ideally "tool0" should be replaced with joint_group_->getEndEffectorName()
+  std::string moveit_tool_link = "tool0";
+
+  if(getFK(joint_point, pose_buffer) == true)  
+  {
+    // Transform Target/Goal Point from moveit_end_link to custom tool frame
+    pose = transformPoseBetweenFrames(pose_buffer, moveit_tool_link,
```
shaun-edwards: This method throws...are you handling that?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -219,16 +217,115 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+  {
+    // Transform Target/Goal Point from moveit_end_link to custom tool frame
+    pose = transformPoseBetweenFrames(pose_buffer, moveit_tool_link, 
+                                      custom_tool_frame);
+
+    return true;
+  }                        
+  else  
+  {
+    return false;
```
shaun-edwards: If you return a failure, you should log an error.  The only exception is if the method is expected to return false (i.e. like a reach check).

moveit_simple/test/moveit_simple.cpp
```diff
@@ -453,6 +453,57 @@ TEST_F(UserRobotTest, kinematics)
 }
 
 
+TEST_F(UserRobotTest, custom_tool_link)
```
shaun-edwards, line 460: This test looks like a copy and paste of the test above.  Is there a better way to reuse code?  Gtest supports scoped tests (i.e. extracting tests into functions) or test fixtures.

moveit_simple/test/resources/kuka_kr210/kuka_custom_tool.urdf
```diff
@@ -0,0 +1,43 @@
...
+
+  <link name="tool_custom"/>
+  
+  <joint name="tool0-tool_custom" type="fixed">
+    <parent link="tool0"/>
+    <child link="tool_custom"/>
+    <origin rpy="3.142 -0.964 3.142" xyz="1.0 3.0 1.5"/>
+  </joint>
+
+  <!-- Adding some waypoints for testing purposes -->
```
shaun-edwards: Please delete commented out code.

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -179,6 +178,21 @@ class Robot
...
+   * @brief This function supports custom tool frame for solving IK.
+   * @param pose - desired pose
+   * @param custom_tool_frame - custom frame id
+   * @param timeout - ik solver timeout
+   * @param attempts - number of IK solve attem
+   * @param seed - ik seed
+   * @param joint_point - joint solution for pose
+   * @return true if joint solution found
+   */
+  bool getJointSolution(const Eigen::Affine3d &pose, const std::string& custom_tool_frame,
```
nsnitish: Are you overloading the original getJointSolution and getPose?

moveit_simple/CMakeLists.txt
MauleshSTrivedi: Oh I changed this but did not test for a while.. wanted to make sure I remembered to check this in case something broke.. I'll delete the comment..

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: I'll make sure to fix this issue when I send out the PR for overloading addTrajPoint(...) and the execute(...) functions.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: ok

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: yes. and I'm gonna overload the addTrajPoint(...) and the execute(...) methods as well and test it on our waypoints in the URDF.

moveit_simple/test/moveit_simple.cpp
MauleshSTrivedi: I can look into that. I've also added tests for execute(...) but I haven't commited those changes yet.

MauleshSTrivedi:
@shaun-edwards Could you please review this update. Thanks.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -230,6 +324,48 @@ bool Robot::getPose(const std::vector<double> & joint_point,
 
 
 
+bool Robot::getPose(const std::vector<double> & joint_point,
+                    const std::string& custom_tool_frame,
+                    Eigen::Affine3d & pose) const
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+  
+  bool GET_POSE_FLAG = false;
```
shaun-edwards: Variable names should be `lower_case`.  Don't append the word `flag`.   It's redundant with the data type, which the IDE displays for you.

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: Am I missing the overloads for `addTrajPoint`?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -220,8 +220,102 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
...
+      {
+        ROS_INFO_STREAM("Empty seed passed to getJointSolution, using current state");
+        local_seed =  getJointState();
+      }
+
+      GET_JOINTS_FLAG = getIK(custom_frame_goal_pose, local_seed, joint_point, timeout);
+  }
+  catch(tf2::TransformException &ex)
+  {
+      ROS_WARN_STREAM("getJointSolution failed for arbitrary pose in Frame[" <<
```
shaun-edwards: I know you don't have to explicitly set the flag since it is already initilaized above, but it helps readability if you set it to `false` in this block.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -220,8 +220,102 @@ bool Robot::getJointSolution(const Eigen::Affine3d &pose, double timeout,
 }
 
 
+
+bool Robot::getJointSolution(const Eigen::Affine3d &pose, const std::string& custom_tool_frame,
+                             double timeout, const std::vector<double> & seed,
+                             std::vector<double> & joint_point) const
+{
+  bool GET_JOINTS_FLAG = false;
```
shaun-edwards: Variable names should be `lower_case`.  Don't append the word `flag`.   It's redundant with the data type, which the IDE displays for you.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -230,6 +324,48 @@ bool Robot::getPose(const std::vector<double> & joint_point,
...
+
+        // Transform Target/Goal Point from moveit_end_link to custom tool frame
+        pose = transformPoseBetweenFrames(pose_buffer, moveit_tool_link, 
+                                          custom_tool_frame);
+
+        GET_POSE_FLAG = true;
+    }
+    catch(tf2::TransformException &ex)
+    {
+        ROS_WARN_STREAM("getPose failed for arbitrary Joint Point in Frame[" <<
```
shaun-edwards: See note above about setting return explicitly.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -230,6 +324,48 @@ bool Robot::getPose(const std::vector<double> & joint_point,
...
+    catch(tf2::TransformException &ex)
+    {
+        ROS_WARN_STREAM("getPose failed for arbitrary Joint Point in Frame[" << 
+            custom_tool_frame << "]: " << ex.what());
+    }    
+  }                        
+  else  
+  {
+    ROS_ERROR_STREAM("Robot State failed to find FK for given joint point");
+  }
```
shaun-edwards: See note above about setting return result explicitly.

shaun-edwards (changes requested):
Almost there...see comments below.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Oh yeah, I have the overloads ready but the tests have been commented out due to another issue somewhere.. I'll just send them out shortly.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Ok sure.

MauleshSTrivedi:
@shaun-edwards Could you please review this latest PR. Thank you.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -165,6 +163,47 @@ class Robot
                     const unsigned int num_steps = 0,
                     const std::string & point_name = std::string());
 
+
+  /**
+   * @brief addTrajPoint - add point to trajectory.
+   * @brief This function supports custom tool frame for adding traj points.
+   * @param traj_name - name of trajectory buffer to add point to
+   * @param point_name - name of point to add
+   * @param custom_tool_frame - frame (must be a TF accessible frame) in which pose is defined
```
shaun-edwards: Change this argument to `tool_name`.  Then add an issue for expanding the source of the `tool_name` transforms (i.e. just like a point can be taken from multiple sources.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -165,6 +163,47 @@ class Robot
...
+                    const std::string & custom_tool_frame, 
+                    double time, const InterpolationType & type = interpolation_type::JOINT,
+                    const unsigned int num_steps = 0);
+  /**
+   * @brief Add trajectory point to motion buffer.
+   * @brief This function supports custom tool frame for adding traj points.
+   * @param traj_name - name of trajectory buffer to add point to
+   * @param pose - pose of point to add
+   * @param pose_frame - frame (must be a TF accessible frame) in which pose is defined
+   * @param custom_tool_frame - frame (must be a TF accessible frame) in which pose is defined
```
shaun-edwards: See note about `custom_tool_frame` above.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -179,12 +218,39 @@ class Robot
                         std::vector<double> & joint_point) const;
 
   /**
+   * @brief getJointSolution returns joint solution for cartesian pose.
+   * @brief This function supports custom tool frame for solving IK.
+   * @param pose - desired pose
+   * @param custom_tool_frame - frame (must be a TF accessible frame) in which pose is defined
```
shaun-edwards: See note about `custom_tool_frame` above.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -179,12 +218,39 @@ class Robot
...
+  /**
    * @brief getPose finds cartesian pose for the given joint positions.
+   * @brief This function supports custom tool frame for solving FK.
    * @param joint_point - joint positions
    * @param pose - pose corresponding to joint_pose
+   * @param custom_tool_frame - frame (must be a TF accessible frame) in which pose is defined
    * @return true if pose is found
    */
   bool getPose(const std::vector<double> & joint_point,
+             const std::string& custom_tool_frame,
```
shaun-edwards: See note about `custom_tool_frame` above.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -241,6 +307,17 @@ class Robot
   Eigen::Affine3d transformToBase(const Eigen::Affine3d &in,
                                          const std::string &in_frame) const;
 
+    /**
+   * @brief customToolFrameTF transforms custom frame to moveit_end_link.
```
shaun-edwards: Description no longer matches method.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -324,7 +401,12 @@ class Robot
                                Eigen::Affine3d &pose) const;
 
   std::unique_ptr<TrajectoryPoint> lookupTrajectoryPoint(const std::string & name,
-                                                              double time) const;
+                                                         double time) const;
+
+  std::unique_ptr<TrajectoryPoint> lookupTrajectoryPoint(const std::string & traj_name,
+                                                         const std::string & name,
+                                                         const std::string & custom_tool_frame,
```
shaun-edwards: See note about `custom_tool_frame`

moveit_simple/src/moveit_simple.cpp
```diff
@@ -143,7 +187,82 @@ void Robot::addTrajPoint(const std::string & traj_name, const std::string & poin
...
+    {
+      ROS_INFO_STREAM("Looked up cartesian target [" << point_name << 
+                      "] from robot model: ");
+
+      // We take the Origin of our reference frame in consideration as the "known pose"
+      // In this case our reference frame resolves to the variable "point_name"
+      Eigen::Affine3d pose_eigen_buffer;
+      geometry_msgs::Pose pose_buffer;
+
+      pose_buffer.position.x = 0.000;
```
shaun-edwards, line 236: Don't change anything, but the default constructors should make initialize these values.  There isn't a need to do this (but it doesn't hurt, so leave it in).

moveit_simple/test/moveit_simple.cpp
```diff
@@ -453,6 +457,93 @@ TEST_F(UserRobotTest, kinematics)
...
+  EXPECT_TRUE(pose1.isApprox(pose2,1e-3));
+  EXPECT_TRUE(pose1.isApprox(pose3,1e-3));
+
+  // Testing addTrajPoint() with custom_tool_frame as eef
+  Eigen::Affine3d pose_msg_to_eigen;
+  geometry_msgs::Pose pose_buffer;
+
+  // We take the Origin of our reference frame in consideration as the "known pose"
+  // In this case our reference frame resolves to the variable "point_name"
+  pose_buffer.position.x = 0.000;
```
shaun-edwards, line 517: Have you verified that this works if the pose_buffer is not at the origin?

shaun-edwards (changes requested):
Some minor changes...almost there!

moveit_simple/test/moveit_simple.cpp
MauleshSTrivedi: @shaun-edwards yes, I verified that it works ok.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: my bad, I'll fix it.

MauleshSTrivedi:
@shaun-edwards Final update for Issue #20.. Lemme know if this looks good to you.

shaun-edwards (approved):
Thanks for the hard works...congrats on your first PR....the rest should be easier from here on out!

MauleshSTrivedi:
@shaun-edwards Thanks a lot Shaun! Appreciate it.. :D

moveit_simple/include/moveit_simple/moveit_simple.h
nsnitish: @shaun-edwards ! Will there be any significant impact if we add the default argument(for custom_tool_frame) in the original getJointSolution method instead of overloading the whole method?

Issue #8, opened 2017-05-05
Motion timing (%speed) should be a dynamic reconfigure parameter


shaun-edwards:
It would be convenient if `moveit_simple` had a speed "knob" that could be used to speed up or slow down motion timing (from the nominal).

shaun-edwards:
@geoffreychiou, could you take a look at this.

shaun-edwards:
@MauleshSTrivedi, here's another one that you can look at if you are waiting on me.  @geoffreychiou started a PR a while back, but it's really out of date.  Have look.  You can try to update the PR or start a new one (you choice).

MauleshSTrivedi:
@shaun-edwards I will have a loot at it.

PR #29, opened 2017-10-25
Moveit simple Issue #8

MauleshSTrivedi:
Added functionality for dynamically re-configuring the trajectory executing speed using rosparam_handler.


shaun-edwards:
Relies on #20.

moveit_simple/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 2.8.3)
 project(moveit_simple)
 find_package(catkin REQUIRED COMPONENTS
   roscpp
+  rosparam_handler
+  dynamic_reconfigure
```
shaun-edwards: Do you need to depend on both rosparam_handler and dynamic_reconfigure?  One should include the other.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -388,19 +473,54 @@ class OnlineRobot : public Robot
...
+   * prior calls to "addTrajPoint".
+   * @param goal - Joint trajectory goal
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
+   * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
+   * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
+   * in Collision with itself or environment)
+   */
+  void planOnly(const std::string traj_name,
```
shaun-edwards: Change name to `plan`....the Only modify does not add anything.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -388,19 +473,54 @@ class OnlineRobot : public Robot
...
+   * @return
+   */
+  void setSpeedModifier(double speed_modifier);
+
+  /**
+   * @brief setSpeedModifier - Getter method for the execution speed modifier of the 
+   * execute method.
+   * @return speed_modifier_
+   */
+  double getSpeedModifier(void);
```
shaun-edwards: Should be a `const` method.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -410,6 +530,14 @@ class OnlineRobot : public Robot
 
   // ROS objects
   ros::Subscriber j_state_sub_;
+
+  // Dynamic Reconfigure
+  double speed_modifier_;
```
shaun-edwards: I know this is based on my original code, but the terminology and sematics of this are bad.  Can we change it to be `speed_percent` with values from 0->1.0?

shaun-edwards (changes requested):
Please rebase after #20 has been merged.  Also, you are using dynamic reconfigure and the param handler...you should only be using the later.  Please remove references to dynamic reconfigure as these are not needed.

shaun-edwards:
There's too many changes from #20...I will review again once #20 has been merged and this branch has been rebased on indigo-devel.

moveit_simple/CMakeLists.txt
MauleshSTrivedi: I initially ran the tests with just dynamic_reconfigure but later ported to rosparam_handler.. I will remove the necessary dependencies.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: ok

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: ok, sure, I can do that.

moveit_simple/CMakeLists.txt
MauleshSTrivedi: yes, so apparently dynamic_reconfigure needs to be explicitly added otherwise we get dependency issues.

MauleshSTrivedi:
@shaun-edwards Could you please review this. Thanks a lot.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * in Collision with itself or environment)
+   */
+  void executeKnownPlan(const std::string traj_name, 
+                   control_msgs::FollowJointTrajectoryGoal & goal,
+                   bool collision_check = false);
+
+
+  /**
+   * @brief plan out a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
```
shaun-edwards: Why is the trajectory name required?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * prior calls to "addTrajPoint".
+   * @param goal - Joint trajectory goal
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
```
shaun-edwards: Can this still happen if we pass in a joint trajectory?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+  void executeKnownPlan(const std::string traj_name, 
+                   control_msgs::FollowJointTrajectoryGoal & goal,
+                   bool collision_check = false);
+
+
+  /**
+   * @brief plan out a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param goal - Joint trajectory goal
```
shaun-edwards: Using this type leaks ROS types into moveit_simple.  Can we use out own data types instead?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
+   * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
+   * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
+   * in Collision with itself or environment)
+   */
+  void plan(const std::string traj_name,
```
shaun-edwards: Why not return the plan?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
+   * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
+   * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
+   * in Collision with itself or environment)
+   */
+  void plan(const std::string traj_name, 
+          control_msgs::FollowJointTrajectoryGoal & goal,
+          ros::Duration & traj_time,
```
shaun-edwards: Why is this required?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * @param goal - Joint trajectory goal
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
+   * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
+   * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
+   * in Collision with itself or environment)
+   */
+  void plan(const std::string traj_name, 
+          control_msgs::FollowJointTrajectoryGoal & goal,
```
shaun-edwards: See comment above about leaking ROS types.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
   */
   virtual std::vector<double> getJointState(void) const;
 
+  /**
+   * @brief setSpeedModifier - Setter method for the execution speed modifier of the 
+   * execute method.
+   * @param speed_modifier
+   * @return
+   */
+  void setSpeedModifier(double speed_modifier);
```
shaun-edwards: The function name and member variable should have the same name (or about).  This makes it easier to correlate dynamic reconfigure parameters with member variables.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * prior calls to "addTrajPoint".
+   * @param goal - Joint trajectory goal which is a known 'Plan'
+   * @param collision_check - bool to turn check for collision on\off
+   * @throws <moveit_simple::ExecutionFailureException> (Execution failure)
+   * @throws <moveit_simple::IKFailException> (Conversion to joint trajectory failed)
+   * @throws <std::invalid_argument> (Trajectory "traj_name" not found)
+   * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
+   * in Collision with itself or environment)
+   */
+  void executeKnownPlan(const std::string traj_name,
```
shaun-edwards: As opposed to an unknown plan?  The method signature should be different for this method, allowing you to use the same `execute` name.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -465,19 +468,71 @@ class OnlineRobot : public Robot
...
+   * @return
+   */
+  void setSpeedModifier(double speed_modifier);
+
+  /**
+   * @brief setSpeedModifier - Getter method for the execution speed modifier of the 
+   * execute method.
+   * @return speed_modifier_
+   */
+  double getSpeedModifier(void) const;
```
shaun-edwards: See comment above about naming.

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: Should the `plan` method be a part of `OnlineRobot` or the base class?

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: If you move Plan into the base class, then this member method and dynamic reconfigure should be moved to the base class.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -640,6 +720,55 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+
+
+void OnlineRobot::plan(const std::string traj_name, 
+                       control_msgs::FollowJointTrajectoryGoal & goal,
+                       ros::Duration & traj_time,
+                       bool collision_check)
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+
+  const double TIMEOUT_SCALE = 1.25;  //scales time to wait for action timeout.
```
shaun-edwards: This variable is no longer needed.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -659,6 +788,46 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
 
 
 
+void OnlineRobot::reconfigureRequest(moveit_simple_dynamic_reconfigure_Config &config, uint32_t level)
+{
+  params_.fromConfig(config);
+  if (params_.speed_percent > 0.0)
```
shaun-edwards: What do you do if this isn't true.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -659,6 +788,46 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+  {
+    setSpeedModifier(params_.speed_percent);
+  }
+}
+
+
+
+void OnlineRobot::setSpeedModifier(double speed_percent)
+{
+  if (speed_percent <= 1.0 && speed_percent > 0.0)
```
shaun-edwards: You should use constants.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -659,6 +788,46 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+  {
+    speed_modifier_ = speed_percent;
+  }
+  else if((speed_percent > 1.0))  
+  {
+    speed_modifier_ = max_speed;
```
shaun-edwards: Don't declare constant member varriables unless they are used in multiple methods.  Otherwise just create them in the method in which they are used.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -659,6 +788,46 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+    ROS_WARN_STREAM("Clamping Speed from " << speed_modifier_ <<
+                    " to max_speed: " << max_speed);
+  }
+  else if((speed_percent <= 0.0))  
+  {
+    speed_modifier_ = min_speed;
```
shaun-edwards: See comment above.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -659,6 +788,46 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+  {
+    speed_modifier_ = speed_percent;
+  }
+  else if((speed_percent > 1.0))  
+  {
+    speed_modifier_ = max_speed;
+    ROS_WARN_STREAM("Clamping Speed from " << speed_modifier_ <<
+                    " to max_speed: " << max_speed);
+  }
+  else if((speed_percent <= 0.0))
```
shaun-edwards: You should always have an `else`....of if you can't get to an else, remove the last `else if` condition.

shaun-edwards (changes requested):
Please address my comments and resubmit.  I didn't tag all the changes...I will wait for you to update the code.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: ok, I've changed the signature.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: Its needed in the call for toJointTrajectory(...) inside plan(...).

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: No it cannot. But Plan returns a jointTrajectory.. the execute(...) function overload takes a jointTrajectory input and executes said traj. And yes, it will NOT throw an IKFailException. Please check out the documentation for that execute(...) function for more details on the exceptions it throws.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: Moved it to base class.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: It was used for some other reason but I forgot to deprecate it. It has been removed now.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: Done.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: Modified it to reflect this case.

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: I was technically sending a reference as an argument. I've modified it so that it returns the plan now.

MauleshSTrivedi: Returning a complex data structure like a plan is not something I would prefer. Instead, we pass an empty plan (type: control_msgs::FollowJointTrajectoryGoal) as a reference argument to the Plan(...) function. This populates the joint trajectory plan which can later be used in the execute(...) function.

MauleshSTrivedi:
@shaun-edwards Could you please review these final changes. Thanks.

moveit_simple/include/moveit_simple/moveit_simple.h
shaun-edwards: Is there a reason not to return the data type? The use of this method will read "better" this way?

MauleshSTrivedi: Yeah, it does read better that way. I guess the main reason is that I usually pass complex data types as references for efficiency reasons. But in this case you are right. The method is going to be exposed at the user's end and not at the developer's end. That supersedes my reason for efficiency. I will make the changes.

gavanderhoorn: @MauleshSTrivedi: wouldn't RVO take care of this?

Also: I would suggest to avoid this sort of premature optimisation. Methods/functions should return proper types, it makes the intent clear. If/when it turns out that returning 'complex types' (debatable for the plan struct) is a bottleneck, *then* it could make sense to pass in reference objects from caller to callee.

MauleshSTrivedi: yeah actually I forgot about RVO.. Instinctively I usually go for references for these types but your'e right, RVO should take care of this.. :)

MauleshSTrivedi:
@shaun-edwards Hey Shaun, updated the PR.. lemme know what you think. :)

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -71,6 +74,8 @@ struct TrajectoryPointInfo {
 };
 typedef std::vector<TrajectoryPointInfo> TrajectoryInfo;
 
+typedef control_msgs::FollowJointTrajectoryGoal JointTrajectoryType;
```
shaun-edwards: This should be a moveit_simple type and not a ROS type (as we discussed).

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,6 +619,50 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
+    goal.trajectory.joint_names = joint_group_->getVariableNames();
+    try
+    {
+      if ( toJointTrajectory(traj_name, goal.trajectory.points, collision_check) )
+      {
+
+        // Modify the speed of execution for the trajectory based off of the speed_modifier_
+        for (std::size_t i = 0; i < goal.trajectory.points.size(); i++)
+        {
+          goal.trajectory.points[i].time_from_start *= (1.0/speed_modifier_);
```
shaun-edwards: The trajectory scaling by the `speed_modifier_` should either occur in the `plan` method or the `execute` method, but not both.  I'd recommend the `plan` method because that was the whole reason for pulling the `speed_modifier_` into the `Robot` base type.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,6 +619,50 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
+  else
+  {
+    ROS_ERROR_STREAM("Trajectoy [" << traj_name << "] not found");
+    throw std::invalid_argument("No trajectory found named " + traj_name);
+  }
+}
+
+
+
 void OnlineRobot::execute(const std::string traj_name, bool collision_check)
```
shaun-edwards: Why doesn't this method call `plan` followed by `execute` with the planned trajectory?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -619,46 +671,157 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
   {
-    ROS_ERROR_STREAM("Trajectoy " << traj_name << " not found");
+    ROS_ERROR_STREAM("Trajectoy [" << traj_name << "] not found");
     throw std::invalid_argument("No trajectory found named " + traj_name);
   }
 }
 
 
 
+void OnlineRobot::execute(JointTrajectoryType & goal,
```
shaun-edwards: The method should simply check all points for collision, then populate the action goal and send.  If there is any collision, then the method should throw.  The structure below does not have to be that complex.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -619,46 +671,157 @@ void OnlineRobot::execute(const std::string traj_name, bool collision_check)
...
+  }
+  else if ((speed_modifier > 1.0))  
+  {
+    speed_modifier_ = 1.0;
+    ROS_WARN_STREAM("Clamping Speed from " << speed_modifier_ <<
+                    " to max_speed: [1.0]");
+  }
+  else
+  {
+    speed_modifier_ = 0.1;
```
shaun-edwards: I recommend a lower limit....like 0.01

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: done

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: I have changed this param, but I would recommend against this.. the simple reason being that a speed of 0.01 is soo soo slow that, more often than not, the program crashes due to some problem in max timeout or thread or something similar.

MauleshSTrivedi:
@shaun-edwards Could you have a look at this update please. Thanks.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -318,6 +341,12 @@ class Robot
                                          const std::string& frame_in,
                                          const std::string& frame_out) const;
 
+  std::vector<moveit_simple::JointTrajectoryPoint> 
+    trajectoryTypeConversionFromROSToMoveItSimple(std::vector<trajectory_msgs::JointTrajectoryPoint> & ROS_joint_trajectory_points)  const;
```
shaun-edwards: Way too long.  The return and parameter types make this obvious.  How about `toJointTrajectoryPoint` and `toFollowJointTrajectoryGoal`.  Because parameters can be different (overrides), but return types cannot, this ensures you will be able to support future conversions.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -383,6 +412,10 @@ class Robot
   bool isReachable(std::unique_ptr<TrajectoryPoint> & point, double timeout,
                    std::vector<double> joint_seed = std::vector<double>() ) const;
 
+  bool jointTrajectoryCollisionCheck(control_msgs::FollowJointTrajectoryGoal & goal,
```
shaun-edwards: This could be shorter...again the parameters indicate this works on a joint trajectory...so no need for an extra long name.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -406,9 +439,22 @@ class Robot
...
                       const std::vector<double> jp2) const;
 
+  /**
+   * @brief toJointTrajPtMsg - Converts native joint point (vector + time) to ROS joint
+   * trajectory point message type.
+   * @param joint_point
+   * @param time
+   * @return
+   */
+  static trajectory_msgs::JointTrajectoryPoint toJointTrajPtMsg(
```
shaun-edwards, line 446: Good naming!

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -406,9 +439,22 @@ class Robot
...
+      const std::vector<double> & joint_point, double time);
```
shaun-edwards, line 447: What is the `time` parameter used for?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,54 +619,233 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
 {
   std::lock_guard<std::recursive_mutex> guard(m_);
 
-  const double TIMEOUT_SCALE = 1.25;  //scales time to wait for action timeout.
-  bool success = false;
   if ( traj_info_map_.count(traj_name) )
   {
-    control_msgs::FollowJointTrajectoryGoal goal;
-    goal.trajectory.joint_names = joint_group_->getVariableNames();
+  	std::vector<trajectory_msgs::JointTrajectoryPoint> ROS_trajectory_points;
```
shaun-edwards, line 629: Correct spacing.

shaun-edwards: naming should be `lower_case`

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,54 +619,233 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
+
+
+void OnlineRobot::execute(std::vector<moveit_simple::JointTrajectoryPoint> & joint_trajectory_points,
+                          bool collision_check)
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+
+  const double TIMEOUT_SCALE = 1.25;  //scales time to wait for action timeout.
+  
+  int collision_points = 0;
```
shaun-edwards: Is this still used?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -383,6 +412,10 @@ class Robot
   bool isReachable(std::unique_ptr<TrajectoryPoint> & point, double timeout,
                    std::vector<double> joint_seed = std::vector<double>() ) const;
 
+  bool jointTrajectoryCollisionCheck(control_msgs::FollowJointTrajectoryGoal & goal,
+                                     int & collision_points,
```
shaun-edwards: This parameter seems redundant with the return value.  Could you just return `0` if not collisions are found?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,54 +619,233 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
+
+  	throw CollisionDetected("Collision detected while interpolating ");
+  }
+}
+
+
+
+std::vector<moveit_simple::JointTrajectoryPoint> Robot::trajectoryTypeConversionFromROSToMoveItSimple(std::vector<trajectory_msgs::JointTrajectoryPoint> & ROS_joint_trajectory_points)  const
+{
+	std::vector<moveit_simple::JointTrajectoryPoint> goal;
```
shaun-edwards: Don't change, but since you know the size of the trajectory you are converting, you may consider calling `reserve` ahead of time for better performance.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -611,54 +619,233 @@ void Robot::clearTrajectory(const::std::string traj_name)
...
+
+bool Robot::jointTrajectoryCollisionCheck(control_msgs::FollowJointTrajectoryGoal & goal,
+										  int & collision_count,
+										  bool collision_check)	
+{
+	bool isStatecolliding = false;
+	collision_count = 0;
+
+	if(collision_check)
+	{
```
shaun-edwards: Correct spacing.  `tabs` should be `2 spaces`.

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: ok

moveit_simple/include/moveit_simple/moveit_simple.h
MauleshSTrivedi: For time from start.

moveit_simple/src/moveit_simple.cpp
MauleshSTrivedi: yes, but only in this method.

MauleshSTrivedi:
@shaun-edwards Could you review these final changes for Issue #8 (Anytime after 15 mins from now -- Travis is still building). Thanks.

shaun-edwards (approved):
Looks good!  Thanks for the effort and patience.

MauleshSTrivedi:
Thanks a lot Shaun. Anytime. :)

PR #34, opened 2017-11-05
Misc build script and pkg manifest fixes

gavanderhoorn:
As per subject.

`run` and `build_depends` can probably be pruned a bit more, but that can be done later.

Compilation is still surprisingly slow on the system that I used. Not sure what is going on there. Perhaps breaking up the single source file into multiple files might help?


shaun-edwards (approved):
Thanks for the contribution.

PR #37, opened 2017-11-22
IKFast integration with collision check for custom tool -- TODO: Add unit tests

MauleshSTrivedi:
First commit. I will be adding the updated URDFs with the unit tests for custom_tool collision check.


MauleshSTrivedi:
@shaun-edwards Could you please review these changes. I have added a separate resources dir as well but I will be pushing those commits once I run some unit tests for a custom tool with geometry. Thanks.

Issue #49, opened 2018-03-31
Unit Tests for Kuka and Motoman Robots are Nearly Identical


geoffreychiou:
We currently run unit tests on two separate robots: the kuka kr210 and motoman mh5. The tests are essentially the same except for some minor differences. 

Instead of separating the tests into two different cpp files with essentially the same code, we should pass the constants in to a single file. This will make it easier to maintain the tests.

geoffreychiou:
Just a reminder to myself to load the kinematics solver from a parameter and run tests for the MH5 on ikfast and LMA.

geoffreychiou:
There is a bug in the UserRobotTest::copy_current_pose test where we get the joint state of the robot right after an execution and compare it to another vector. We need to add a function to compare two vectors with some allowable error (or find one in g_test that does the same).

Issue #48, opened 2018-03-31
Unit Tests for Singularity (isNearSingular() calls) are Disabled for Motoman MH5


geoffreychiou:
The isNearSingular() method call in the motoman mh5 singularity tests are always returning true for whatever joint pose we try. They have been commented out for now.

MauleshSTrivedi:
IsNearSingular(...) has always been really hard to test

PR #50, opened 2018-03-31
Ability to use generic ikfast solution with a robot (Motoman MH5 + tests)

geoffreychiou:
Added the ability to use a generic ikfast solution by calling a different robot constructor.

The ikfast solution is defined from base_link to link_t, and the robot SRDF is defined from base_link to tool_custom.

A couple of TODO's have been logged as issues:

1. The singularity unit tests made calls to the isNearSingular() method, which always returned true for every joint pose I passed in. Tracked in #48.

1. The unit tests for the mh5 were added in as a copy of kuka tests. This is un-maintainable and the actual test functions need to be the same. This is the next highest priority for me, tracked in #49.

@shaun-edwards please review.


geoffreychiou:
Oh yea, still need to test this with RVIZ running.

geoffreychiou:
@shaun-edwards, just added a launch file to view tests in RVIZ while they are running and some installation instructions to the README.

As far as I can tell, the tool_custom frame matches up with the waypoints. Can you please clone my branch and verify?

.travis.rosinstall
```diff
@@ -1,2 +1,4 @@
+- git: {local-name: industrial_experimental, uri: 'https://github.com/ros-industrial/industrial_experimental.git', version: indigo-devel}
```
shaun-edwards, line 1: Do we need `industrial_experimental` and `motoman_experimental`?  In the past, we've embedded the required files/packages in the tests directory.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -349,6 +360,13 @@ class Robot
                          std::vector<trajectory_msgs::JointTrajectoryPoint> & points,
                          bool collision_check = false);
 
+  /**
+   * @brief computeIKTransforms - Checks if the frames passed in for a custom
+   * ikfast solver exist and computes the transform from the last link defined
+   * for the solver kinematics to the tcp link defined in the robot description
+   */
+  bool computeIKTransforms();
```
shaun-edwards: It's hard to tell what this method does from the name.  The description isn't helping me either.  The fact that it returns a bool gives a hint, but I'm still not sure.

moveit_simple/package.xml
```diff
@@ -34,21 +34,22 @@
   <run_depend>eigen</run_depend>
   <run_depend>eigen_conversions</run_depend>
   <run_depend>geometry_msgs</run_depend>
+  <run_depend>industrial_robot_client</run_depend>
```
shaun-edwards, line 37: Why did these change from `test_depend` to `run_depend`

moveit_simple/src/moveit_simple.cpp
```diff
@@ -895,7 +909,30 @@ bool Robot::toJointTrajectory(const std::string traj_name,
   return true;
 }
 
+bool Robot::computeIKTransforms()
```
shaun-edwards: This method wraps an existing method called `knowsFrameTransform`...is there a reason you just didn't re-use the name?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -895,7 +909,30 @@ bool Robot::toJointTrajectory(const std::string traj_name,
...
+{
+  if (!virtual_robot_state_->knowsFrameTransform(ik_base_frame_))
```
shaun-edwards: Throughout this class we use `tf` for getting frame transformations.  I believe this is the better approach because `tf` has access to more transforms than the `robot_state`.  I believe `robot_state` only knows about the frames defined in the URDF.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -895,7 +909,30 @@ bool Robot::toJointTrajectory(const std::string traj_name,
...
+  if (!virtual_robot_state_->knowsFrameTransform(ik_tip_frame_))
+  {
+    ROS_ERROR_STREAM("Unable to find the transformation to frame: "
+      << ik_tip_frame_ << " in group: " << joint_group_->getName()
+      << " IK solutions will not be found!");
+    return false;
+  }
 
+  std::vector<std::string> link_names = joint_group_->getLinkModelNames();
+  tool0_to_tcp_ = virtual_robot_state_->getFrameTransform(link_names.back()).inverse()
```
shaun-edwards: Does this require the frame transform from `tool0_to_tcp_` to be fixed?

I would use another name other than `tool0`.  The name is inconsistent with the rest of the class.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -1206,14 +1243,26 @@ bool Robot::getFK(const std::vector<double> & joint_point,
                   Eigen::Affine3d &pose) const
 {
   virtual_robot_state_->setJointGroupPositions(joint_group_, joint_point);
-  const std::vector<std::string> link_names = joint_group_->getLinkModelNames();
-  const int vc =  (int) virtual_robot_state_->getVariableCount();
-  if ( joint_point.size() == vc)
+
+  if (ik_transforms_found_)
```
shaun-edwards: If this were named `ik_transforms required`, would it make more sense.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -62,7 +60,15 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
   return;
 }
 
+Robot::Robot(const ros::NodeHandle &nh, const std::string &robot_description,
+  const std::string &group_name, const std::string &ik_base_frame,
+  const std::string &ik_tip_frame) : Robot(nh, robot_description, group_name)
+{ 
+  ik_base_frame_ = ik_base_frame;
```
shaun-edwards: Why aren't `ik_base_frame` and `ik_tip_frame` set in the default constructor as the first/last links in the moveit chain.  Not setting class values to appropriate values is a recipe for bugs.  Other developers may not understand that they are conditionally set.  Also, if the values are appropriate no matter how the object is constructed then the class methods can operate on them without checking a condition variable.  This will make your methods more straight-forward.

moveit_simple/src/moveit_simple.cpp
```diff
@@ -1206,14 +1243,26 @@ bool Robot::getFK(const std::vector<double> & joint_point,
...
-  }else{
-    return false;
+  }
+  else
+  {
+    const std::vector<std::string> link_names = joint_group_->getLinkModelNames();
+    const int vc =  (int) virtual_robot_state_->getVariableCount();
+    if (joint_point.size() == vc)
+    {
+      pose = virtual_robot_state_->getFrameTransform(link_names.back());
```
shaun-edwards: Would this bit of logic be needed if `ik_tip_frame` was set in the constructor?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -1232,16 +1281,34 @@ bool Robot::getIK(const Eigen::Affine3d pose, const std::vector<double> & seed,
...
+      virtual_visual_tools_->deleteAllMarkers();
+      virtual_visual_tools_->publishRobotState(virtual_robot_state_, rviz_visual_tools::PURPLE);
+      virtual_visual_tools_->publishContactPoints(*virtual_robot_state_, &(*planning_scene_));  
+      ros::spinOnce();
+      return true;      
+    }
+  }
+  else
+  {
+    if (virtual_robot_state_->setFromIK(joint_group_, pose, attempts, timeout))
```
shaun-edwards: This is a direct copy from above.  Can it be written better?

shaun-edwards (changes requested):

shaun-edwards:
I'm surprised there aren't any changes to the kuka tests?

The main issue is you are treating the iksolver sync issue as a separate condition.  However, whether the iksolvers match the SRDF or not, they all have a base/tip link.  The code may be a lot more straight-foward and easier to maintain if this distinction between sync/not in sycn with the SRDF were not made.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -477,11 +495,15 @@ class Robot
...
   moveit_simple_dynamic_reconfigure_Parameters params_;
-
   dynamic_reconfigure::Server
   <moveit_simple_dynamic_reconfigure_Config> dynamic_reconfig_server_;
+
+  // Kinematics
+  std::string ik_base_frame_;
+  std::string ik_tip_frame_;
+  bool ik_transforms_found_;
+  Eigen::Affine3d tool0_to_tcp_;
```
shaun-edwards: Seems odd that you don't have a similar transform for the base?

moveit_simple/src/moveit_simple.cpp
```diff
@@ -62,7 +60,15 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
 
+Robot::Robot(const ros::NodeHandle &nh, const std::string &robot_description,
+  const std::string &group_name, const std::string &ik_base_frame,
+  const std::string &ik_tip_frame) : Robot(nh, robot_description, group_name)
+{ 
+  ik_base_frame_ = ik_base_frame;
+  ik_tip_frame_ = ik_tip_frame;
 
+  ik_transforms_found_ = this->computeIKTransforms();
+}
```
shaun-edwards: You should output a log message that indicates this special handling of base and tip frames.

shaun-edwards (changes requested):

.travis.rosinstall
geoffreychiou: The ikfast plugin for the mh5 is in motoman_experimental, which depends on industrial_experimental.

moveit_simple/package.xml
geoffreychiou: I ran roslaunch check and it complained.

moveit_simple/include/moveit_simple/moveit_simple.h
geoffreychiou: Renamed to computeIKSolverTransforms() and added a more descriptive description.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Don't need this anymore.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/include/moveit_simple/moveit_simple.h
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Yes.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Nope.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: It does more.

geoffreychiou:
@shaun-edwards can you please review latest changes.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -477,11 +494,15 @@ class Robot
...
-
   moveit_simple_dynamic_reconfigure_Parameters params_;
-
   dynamic_reconfigure::Server
   <moveit_simple_dynamic_reconfigure_Config> dynamic_reconfig_server_;
+
+  // Kinematics
+  std::string ik_base_frame_;
+  std::string ik_tip_frame_;
+  Eigen::Affine3d ik_tip_to_urdf_tip_;
```
shaun-edwards: Shouldn't this bet `srdf` and not `urdf`

moveit_simple/src/moveit_simple.cpp
```diff
@@ -59,10 +56,41 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
+  try
+  {
+    ik_base_frame_ = ik_base_frame;
+    ik_tip_frame_ = ik_tip_frame;    
+    this->computeIKSolverTransforms();
+  }
+  catch (tf2::TransformException &ex)
+  {
+    ROS_ERROR_STREAM("Failed to compute transforms between the base/tip frames defined"
+      << " in the URDF and the base/tip frames defined for the IK solver");
```
shaun-edwards: SRDF

moveit_simple/src/moveit_simple.cpp
```diff
@@ -59,10 +56,41 @@ Robot::Robot(const ros::NodeHandle & nh, const std::string &robot_description,
...
+    throw IKSolverTransformException("Failed to compute transforms between the base/tip" 
+      " frame defined in the URDF and the base/tip frames defined for the IK solver");
```
shaun-edwards: SRDF

moveit_simple/src/moveit_simple.cpp
```diff
@@ -95,11 +123,41 @@ OnlineRobot::OnlineRobot(const ros::NodeHandle & nh,
...
+  try
+  {
+    ik_base_frame_ = joint_group_->getSolverInstance()->getBaseFrame();
+    ik_tip_frame_ = joint_group_->getSolverInstance()->getTipFrame();    
+    this->computeIKSolverTransforms();
+  }
+  catch (tf2::TransformException &ex)
+  {
+    ROS_ERROR_STREAM("Failed to compute transforms between the base/tip frames defined"
+      << " in the URDF and the base/tip frames defined for the IK solver");
```
shaun-edwards: SRDF

moveit_simple/src/moveit_simple.cpp
```diff
@@ -95,11 +123,41 @@ OnlineRobot::OnlineRobot(const ros::NodeHandle & nh,
...
+    throw IKSolverTransformException("Failed to compute transforms between the base/tip" 
+      " frame defined in the URDF and the base/tip frames defined for the IK solver");
```
shaun-edwards: SRDF

moveit_simple/src/moveit_simple.cpp
```diff
@@ -95,11 +123,41 @@ OnlineRobot::OnlineRobot(const ros::NodeHandle & nh,
...
+  }
 
   return;
 }
 
-
+OnlineRobot::OnlineRobot(const ros::NodeHandle &nh, const std::string &robot_description,
```
shaun-edwards, line 129: Why don't you call the base class constructor?

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -136,7 +147,7 @@ class Robot
    * @brief addTrajPoint - add point to trajectory.
    * @param traj_name - name of trajectory buffer to add point to
    * @param point_name - name of point to add
-   * @param time - time from start of trajectory to reach point
+   * @param time - time from sta+rt of trajectory to reach point
```
shaun-edwards: Remove.

moveit_simple/include/moveit_simple/moveit_simple.h
```diff
@@ -87,6 +87,17 @@ class Robot
 public:
    Robot(const ros::NodeHandle & nh, const std::string &robot_description,
                                      const std::string &group_name);
+  
+  /**
+  * @brief Constructor for the case where the IK implementation does not match the
+  * SRDF. For example: IKFast solutions are generally solved for the base_link to
+  * tool0 of a robot, and the robot_description is defined for some world_frame to
+  * some tcp frame.
+  */
```
shaun-edwards, line 96: Add a note that the transforms to the IK frames are assumed to be static.  It would be nice if TF told us if a particular transform was static.  Then the class could automatically adapt to static/non-static transforms.

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/moveit_simple.h
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Because the regular OnlineRobot constructor does a bunch of stuff after calling the base class constructor.

moveit_simple/include/moveit_simple/moveit_simple.h
geoffreychiou: Yes.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/src/moveit_simple.cpp
geoffreychiou: Done.

moveit_simple/include/moveit_simple/moveit_simple.h
geoffreychiou: Done.

geoffreychiou:
@shaun-edwards the requested changes have been made.

geoffreychiou:
@shaun-edwards should we consider leaving ```indigo-devel``` as it was and merging this into a new ```kinetic-devel``` branch?

PR #51, opened 2018-04-27
Fixed Bug in Transforms for Custom IK Solver

geoffreychiou:
Fixed a bug where we didn't account for the case where there is a transform between the root frame of the ik solver and the root frame of the model.

Still need to test this on hardware.

@gomezc please review.

### DIFF
--- a/moveit_simple/src/moveit_simple.cpp
+++ b/moveit_simple/src/moveit_simple.cpp
@@ -340,19 +340,20 @@ std::unique_ptr<TrajectoryPoint> Robot::lookupTrajectoryPoint(const std::string 
   {
     ROS_INFO_STREAM("Looked up named cart target from robot model: " << name);
     virtual_robot_state_->update();  //Updating state for frame tansform below
-    Eigen::Affine3d pose =  virtual_robot_state_->getFrameTransform(name);
+    Eigen::Affine3d ik_base_to_ik_tip =  virtual_robot_state_->getFrameTransform(name);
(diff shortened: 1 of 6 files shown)


moveit_simple/src/moveit_simple.cpp
```diff
@@ -340,19 +340,20 @@ std::unique_ptr<TrajectoryPoint> Robot::lookupTrajectoryPoint(const std::string
   {
     ROS_INFO_STREAM("Looked up named cart target from robot model: " << name);
     virtual_robot_state_->update();  //Updating state for frame tansform below
-    Eigen::Affine3d pose =  virtual_robot_state_->getFrameTransform(name);
+    Eigen::Affine3d ik_base_to_ik_tip =  virtual_robot_state_->getFrameTransform(name);
+    Eigen::Affine3d pose = srdf_base_to_ik_base_ * ik_base_to_ik_tip;
```
gomezc, line 344: I'm not 100% sure about this line where you redefine the pose. It feels like this is a very different definition of what pose used to be, but I'm not familiar with "getFrameTransform" or what uses the pose called by lookupTrajectoryPoint

gomezc:
One comment - all other changes look good.

moveit_simple/src/moveit_simple.cpp
gomezc: talked with @geoffreychiou and found where srdf_base_to_ik_base_ would get set to identity under the condition where base is aligned with world and then this essentially looks the same as before.

gomezc:
+1, @shaun-edwards can you please merge

PR #53, opened 2018-05-15
Added an Overload for isReachable() and isInCollision() Taking in a Named Seed Pose

geoffreychiou:
@gomezc please review.


gomezc:
I agree that I don't like that there is are two similar functions, but I see why it is here. +1 from me

PR #55, opened 2018-05-23
Added isReachableStatic and isInCollisionStatic and transformToBaseStatic

geoffreychiou:
The original functions called transformToBase, which made a tf lookup for every single request causing lots of lag. These functions assume the transform to be static and will only look it up for the first instance.

@joshuaplusone please review.

### DIFF
--- a/moveit_simple/include/moveit_simple/robot.h
+++ b/moveit_simple/include/moveit_simple/robot.h
@@ -70,6 +70,31 @@ public:
   ~Robot();
 
   /**
+   * @brief isInCollision  returns true if pose results in robot config that is
+   * in collision with the environment as defined by the URDF.
+   * @param pose - cartesian pose of the tool to check
+   * @param frame - tool pose relative to frame
+   * @param joint_seed - named seed to use defined in srdf
+   * @param timeout - (optional) timeout for IK
+   * @return
+   */
+  bool isInCollision(const Eigen::Affine3d &pose, const std::string &frame,
+    const std::string &joint_seed, double timeout = 10.0) const;
+
+  /**
+   * @brief isInCollision  returns true if pose results in robot config that is
+   * in collision with the environment as defined by the URDF.
+   * @param pose - cartesian pose of the tool to check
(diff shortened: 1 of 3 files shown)


geoffreychiou:
Setting to WIP as @joshuaplusone has some requested changes.

moveit_simple/test/kuka_kr210.cpp
```diff
@@ -647,15 +647,18 @@ TEST_F(UserRobotTest, copy_current_pose)
   // Before the first execution robot should be at position "home"
   before_execution_1 = robot->getJointState();
   EXPECT_NO_THROW(robot->execute(TRAJECTORY_NAME));
+  ros::Duration(0.5).sleep();
```
shaun-edwards, line 650: Why is this needed?

shaun-edwards (dismissed):

moveit_simple/test/kuka_kr210.cpp
geoffreychiou: I think we had this discussion for the mh5 tests, added something similar [there](https://github.com/plusone-robotics/moveit_simple/blob/kinetic-devel/moveit_simple/test/motoman_mh5.cpp#L750). It was something about the industrial_robot_simlator was still at the end of the execution before the next getJointState() was called, causing the tests to sometimes fail or sometimes pass.

geoffreychiou:
@joshuaplusone refactored as your requested, this is review ready.

moveit_simple/include/moveit_simple/robot.h
```diff
@@ -87,20 +112,21 @@ class Robot
    * @param joint_seed (optional) - seed to use
    * @return
    */
-  bool isInCollision(const Eigen::Affine3d pose, const std::string &frame,
+  bool isInCollision(const Eigen::Affine3d &pose, const std::string &frame,
```
joshuaplusone, line 115: Nice catch.

joshuaplusone (approved):
Looks good to me.

PR #59, opened 2018-07-25
Added non blocking execution call for online robot

geoffreychiou:
TODO

- [x] Add tests for new functionality.
- [x] Make interface less clunky and try to integrate timeout inside of the class.

### DIFF
--- a/moveit_simple/include/moveit_simple/online_robot.h
+++ b/moveit_simple/include/moveit_simple/online_robot.h
@@ -79,6 +79,7 @@ public:
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
    * @param collision_check - bool to turn check for collision on\off
+   * @return Goal successfully set without errors
    */
   bool setExecuteGoal(const std::string &traj_name, bool collision_check = false);
 
@@ -87,6 +88,7 @@ public:
    * @param traj_name - name of trajectory to be executed (must be filled with
    * prior calls to "addTrajPoint".
    * @param collision_check - bool to turn check for collision on\off
+   * @return Goal successfully set without errors
    */
   bool setExecuteGoal(const std::vector<moveit_simple::JointTrajectoryPoint> &goal_points, bool collision_check = false);
 
--- a/moveit_simple/include/moveit_simple/online_robot.h
+++ b/moveit_simple/include/moveit_simple/online_robot.h
@@ -72,8 +72,21 @@ public:
    * @throws <moveit_simple::CollisionDetected> (One of interpolated point is
    * in Collision with itself or environment)
    */
-  void execute(std::vector<moveit_simple::JointTrajectoryPoint> &goal,
-    bool collision_check = false);
+  void execute(std::vector<moveit_simple::JointTrajectoryPoint> &goal, bool collision_check = false);
+
+  bool setExecuteGoal(const std::string &traj_name, bool collision_check = false);
+
+  bool setExecuteGoal(const std::vector<moveit_simple::JointTrajectoryPoint> &goal_points, bool collision_check = false);
+
+  void startExecution();
+
+  void stopExecution();
+
+  bool isExecuting();
+
+  bool checkGoalInCollision();
+
+  bool getExecutionTimeout(ros::Duration &timeout);
 
   /**
   * @brief getJointState - Returns a vector<double> of the
@@ -91,6 +104,10 @@ protected:
   moveit_visual_tools::MoveItVisualToolsPtr online_visual_tools_;
   actionlib::SimpleActionClient<control_msgs::FollowJointTrajectoryAction> action_;
   ros::Subscriber j_state_sub_;
+
+  control_msgs::FollowJointTrajectoryGoal execution_goal_;
+  bool execute_collision_check_ = false;
+  int execute_collision_points_ = 0;
 };
 } // namespace moveit_simple
 #endif // ONLINE_ROBOT_H
--- a/moveit_simple/include/moveit_simple/online_robot.h
+++ b/moveit_simple/include/moveit_simple/online_robot.h
@@ -74,18 +74,45 @@ public:
    */
   void execute(std::vector<moveit_simple::JointTrajectoryPoint> &goal, bool collision_check = false);
 
+  /**
+   * @brief Set the goal to execute a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param collision_check - bool to turn check for collision on\off
+   */
   bool setExecuteGoal(const std::string &traj_name, bool collision_check = false);
 
+  /**
+   * @brief Set the goal to execute a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param collision_check - bool to turn check for collision on\off
+   */
   bool setExecuteGoal(const std::vector<moveit_simple::JointTrajectoryPoint> &goal_points, bool collision_check = false);
 
+  /**
+   * @brief Executes the given planned trajectory
+   */
   void startExecution();
 
+  /**
+   * @brief Stops executing the given planned trajectory
+   */
   void stopExecution();
 
+  /**
+   * @brief Checks if the given planned trajectory is in pending / active state
+   */
   bool isExecuting();
 
+  /**
+   * @brief Checks if the given planned trajectory goal is in collision
+   */
   bool checkGoalInCollision();
 
+  /**
+   * @brief Gets the timeout for the given planned trajectory
+   */
   bool getExecutionTimeout(ros::Duration &timeout);
 
   /**
--- a/moveit_simple/include/moveit_simple/online_robot.h
+++ b/moveit_simple/include/moveit_simple/online_robot.h
@@ -108,12 +108,12 @@ public:
   /**
    * @brief Checks if the given planned trajectory goal is in collision
    */
-  bool checkGoalInCollision();
+  bool isGoalInCollision();
 
   /**
    * @brief Checks if the given planned trajectory goal is stopped
    */
-  bool checkExecutionStopped();
+  bool isExecutionStopped();
 
   /**
    * @brief Gets the timeout for the given planned trajectory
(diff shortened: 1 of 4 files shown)


moveit_simple/include/moveit_simple/online_robot.h
```diff
@@ -72,8 +72,53 @@ class OnlineRobot : public Robot
...
-    bool collision_check = false);
+  void execute(std::vector<moveit_simple::JointTrajectoryPoint> &goal, bool collision_check = false);
+
+  /**
+   * @brief Set the goal to execute a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param collision_check - bool to turn check for collision on\off
+   */
+  bool setExecuteGoal(const std::string &traj_name, bool collision_check = false);
```
shaun-edwards: What does the return mean?

moveit_simple/include/moveit_simple/online_robot.h
```diff
@@ -72,8 +72,53 @@ class OnlineRobot : public Robot
...
+
+  /**
+   * @brief Set the goal to execute a given trajectory
+   * @param traj_name - name of trajectory to be executed (must be filled with
+   * prior calls to "addTrajPoint".
+   * @param collision_check - bool to turn check for collision on\off
+   */
+  bool setExecuteGoal(const std::vector<moveit_simple::JointTrajectoryPoint> &goal_points, bool collision_check = false);
```
shaun-edwards: What does the return mean?

moveit_simple/include/moveit_simple/online_robot.h
shaun-edwards: Can you think of an application where you would call `setExecuteGoal` and not immediately call `startExecution`?  Perhaps these two methods are always called together and should be combined into a `beginExecute` method?

moveit_simple/include/moveit_simple/online_robot.h
```diff
@@ -72,8 +72,53 @@ class OnlineRobot : public Robot
...
+
+  /**
+   * @brief Checks if the given planned trajectory is in pending / active state
+   */
+  bool isExecuting();
+
+  /**
+   * @brief Checks if the given planned trajectory goal is in collision
+   */
+  bool checkGoalInCollision();
```
shaun-edwards: Should this be named `isGoalInCollision`?

moveit_simple/include/moveit_simple/online_robot.h
```diff
@@ -72,8 +72,53 @@ class OnlineRobot : public Robot
...
+
+  /**
+   * @brief Checks if the given planned trajectory goal is stopped
+   */
+  bool checkExecutionStopped();
```
shaun-edwards: Does this method check if the execution is DONE or STOPPED (or both)?

shaun-edwards (changes requested):
A few questions.

shaun-edwards:
@geoffreychiou, please add appropriate tests for this new functionality.  

At a high level, it feels like you added more functionality than was needed to achieve our use case.   What else is required beyond the code snippet below

```

robot.beginExecute("some_traj");
while (robot.isExecuting())
{
  //check OK conditions
  if (not_ok)
  {
    robot.abortExecution()
  }
}
```

moveit_simple/include/moveit_simple/online_robot.h
geoffreychiou: > What does the return mean?

The return is a boolean value, which in c++ is true or false. It means the goal was successfully set.

> Can you think of an application where you would call setExecuteGoal and not immediately call startExecution?

Yes.  In our case we calculate a timeout to cancel the execution in case it is stuck in isExecuting() but the trajectory isn't supposed to take that long.

moveit_simple/include/moveit_simple/online_robot.h
geoffreychiou: It checks for the following:

```
action_state == actionlib::SimpleClientGoalState::RECALLED 
action_state == actionlib::SimpleClientGoalState::PREEMPTED
action_state == actionlib::SimpleClientGoalState::ABORTED
```

geoffreychiou:
> What else is required beyond the code snippet below

Beyond that, we check for the timeout of the execution locally and just because we call stopExecution() doesn't mean the execution stops immediately, so if I were to add another motion right after it causes problems.

moveit_simple/include/moveit_simple/online_robot.h
geoffreychiou: It can be.

moveit_simple/include/moveit_simple/online_robot.h
shaun-edwards: Do you need that functionality?  The current `execute` method does not currently allow for this.

geoffreychiou: The current execute function has a timeout.

moveit_simple/include/moveit_simple/online_robot.h
shaun-edwards: Ok...should this be renamed `is...`?

geoffreychiou: https://github.com/plusone-robotics/moveit_simple/pull/59/commits/473080e0f170225c64e3544b42dd63f44dc794b4

shaun-edwards:
What is the use case for `checkGoalInCollision`, if it isn't needed in the code snippet above.

geoffreychiou:
> What is the use case for checkGoalInCollision, if it isn't needed in the code snippet above.

It is used as privately, but I left it as public in case someone wanted to use it.

moveit_simple/include/moveit_simple/online_robot.h
shaun-edwards: Ok...but can you make the interface cleaner, less clunky.  Could the class take care of checking for timed out motion in some way?

geoffreychiou: Yea i'll keep that in mind, this is a WIP so I just slapped together minimum amount of code to get it working.

moveit_simple/include/moveit_simple/online_robot.h
geoffreychiou: Added a note to the documentation.

moveit_simple/include/moveit_simple/online_robot.h
```diff
@@ -91,6 +138,9 @@ class OnlineRobot : public Robot
   moveit_visual_tools::MoveItVisualToolsPtr online_visual_tools_;
   actionlib::SimpleActionClient<control_msgs::FollowJointTrajectoryAction> action_;
   ros::Subscriber j_state_sub_;
+
+  control_msgs::FollowJointTrajectoryGoal execution_goal_;
+  bool execute_collision_check_ = false;
```
shaun-edwards: This seems like an odd flag to keep around.  Does it need to be a class variable (is it used by multiple class members, does it's value need to be persistent beyond function calls).

This is likely to cause bugs in the future.

moveit_simple/test/motoman_mh5.cpp
```diff
@@ -864,4 +864,34 @@ TEST_F(UserRobotTest, singularity)
...
+  EXPECT_NO_THROW(user_robot->addTrajPoint(TRAJECTORY_NAME, "wp2", 3.0));
+  EXPECT_NO_THROW(user_robot->addTrajPoint(TRAJECTORY_NAME, "wp3", 4.0, joint));
+
+  EXPECT_TRUE(user_robot->setExecuteGoal(TRAJECTORY_NAME));
+
+  user_robot->startExecution();
+
+  ros::Duration timeout;
+  EXPECT_TRUE(user_robot->getExecutionTimeout(timeout));
+  EXPECT_FALSE(user_robot->isGoalInCollision());
```
shaun-edwards: Why do we need to check this?  Shouldn't `setExecuteGoal` fail?

geoffreychiou: What if someone modifies setExecuteGoal and suddenly it doesn't fail?

moveit_simple/include/moveit_simple/online_robot.h
geoffreychiou: > Does it need to be a class variable (is it used by multiple class members, does it's value need to be persistent beyond function calls).

Yes, it is used by multiple class members. The value is persistent between goals, so if a new goal is set the value gets changed.

geoffreychiou:
@joshuaplusone please review.

moveit_simple/src/online_robot.cpp
```diff
@@ -145,6 +145,130 @@ void OnlineRobot::execute(std::vector<moveit_simple::JointTrajectoryPoint> &join
...
+  }
+}
+
+void OnlineRobot::stopExecution()
+{
+  std::lock_guard<std::recursive_mutex> guard(m_);
+  action_.cancelGoal();
+}
+
+bool OnlineRobot::isExecuting()
```
joshuaplusone: Who's responsibility is it to check isExecuting/isGoalInCollision/isExecution stopped.  My immediate reaction is that these should be private concerns.

moveit_simple/src/online_robot.cpp
```diff
@@ -145,6 +145,130 @@ void OnlineRobot::execute(std::vector<moveit_simple::JointTrajectoryPoint> &join
...
+  {
+    return true;
+  }
+  else
+  {
+    return false;
+  }
+}
+
+bool OnlineRobot::getExecutionTimeout(ros::Duration &timeout)
```
joshuaplusone: I also believe this should be a private concern.

geoffreychiou:
@joshuaplusone I made the changes you requested.

joshuaplusone (approved):

moveit_simple/src/online_robot.cpp
geoffreychiou: This no longer exists.

moveit_simple/src/online_robot.cpp
geoffreychiou: All we have now is the isExecuting() method, and it is the user's responsibility to check. If they didn't, they would just call execute and the code would just move on while the robot is moving in the background.

PR #73, opened 2019-02-10
WIP: Add trajectory training service interface

aaronplusone:
@shaun-edwards please review. WIP pending functional testing.

Extends lookupTrajectoryPoint to also check for point info from a **TrajectoryTrainer** service, and adds a new TrajectoryPoint type **CombinedTrajectoryPoint**.

Trainer service definition and CombinedJointPoint message types are defined in new moveit_simple_msgs package.

### DIFF
--- a/moveit_simple/include/moveit_simple/point_types.h
+++ b/moveit_simple/include/moveit_simple/point_types.h
@@ -33,6 +33,7 @@ class Robot;
 class TrajectoryPoint;
 class JointTrajectoryPoint;
 class CartTrajectoryPoint;
+class CombinedTrajectoryPoint;
 
 namespace interpolation_type
 {
@@ -81,8 +82,7 @@ protected:
   {
     UNKNOWN = 0,
     JOINT = 1,
-    CARTESIAN = 2,
-    COMBINED = 3
+    CARTESIAN = 2
   };
 
   TrajectoryPoint(std::string name, double t, PointType type, JointLockOptions options = JointLockOptions::LOCK_NONE)
@@ -110,6 +110,7 @@ protected:
 class JointTrajectoryPoint : public TrajectoryPoint
 {
 public:
+  friend class CombinedTrajectoryPoint;
   JointTrajectoryPoint() : TrajectoryPoint("", 0.0, PointType::JOINT) { }
   
   virtual ~JointTrajectoryPoint() { }
@@ -136,6 +137,7 @@ private:
 class CartTrajectoryPoint : public TrajectoryPoint
 {
 public:
+  friend class CombinedTrajectoryPoint;
   CartTrajectoryPoint() : TrajectoryPoint("", 0.0, PointType::CARTESIAN) { }
 
   CartTrajectoryPoint(const Eigen::Affine3d pose, const double t, std::string name = std::string(),
@@ -162,11 +164,24 @@ private:
 class CombinedTrajectoryPoint : public TrajectoryPoint
 {
 public:
-  CombinedTrajectoryPoint() : TrajectoryPoint("", 0.0, PointType::COMBINED) { }
+  enum PointPreference
+  {
+    CARTESIAN = 0,
+    JOINT = 1
+  };
 
-  CombinedTrajectoryPoint(std::vector<double> &joint_point, const Eigen::Affine3d pose, double t,
-                          std::string name = std::string(), JointLockOptions options = JointLockOptions::LOCK_NONE)
-    : TrajectoryPoint(name, t, PointType::COMBINED, options), options_(options), joint_point_(joint_point), pose_(pose)
+  CombinedTrajectoryPoint() : TrajectoryPoint("", 0.0, PointType::JOINT)
+  {
+  }
+
+  CombinedTrajectoryPoint(std::vector<double> &joint_point, const Eigen::Affine3d pose, double t, double tol,
+                          std::string name = std::string(), JointLockOptions options = JointLockOptions::LOCK_NONE,
+                          PointPreference type = PointPreference::JOINT, double timeout = 5.0)
+    : TrajectoryPoint(name, t, (type == PointPreference::JOINT) ? PointType::JOINT : PointType::CARTESIAN, options)
+    , options_(options)
+    , joint_point_(joint_point)
+    , pose_(pose)
+    , tol_(tol)
   {
   }
 
@@ -187,19 +202,28 @@ public:
     return joint_point_;
   }
 
+  const double &timeout() const
+  {
+    return timeout_;
+  }
+
 protected:
+  
   virtual std::unique_ptr<JointTrajectoryPoint>
   toJointTrajPoint(const Robot &robot, double timeout, const std::vector<double> &seed,
                    JointLockOptions options = JointLockOptions::LOCK_NONE) const;
 
   virtual std::unique_ptr<CartTrajectoryPoint> toCartTrajPoint(const Robot &robot) const;
 
-  bool compareJointAndCart(const Robot &robot, double timeout, double tolerance);
+  bool compareJointAndCart(const Robot &robot, double timeout) const;
 
 private:
+
   JointLockOptions options_;
   std::vector<double> joint_point_;
   Eigen::Affine3d pose_;
+  double tol_;
+  double timeout_;
 };
 } // namespace moveit_simple
 #endif // POINT_TYPES_H
--- a/moveit_simple/include/moveit_simple/point_types.h
+++ b/moveit_simple/include/moveit_simple/point_types.h
@@ -175,8 +175,8 @@ public:
   }
 
   CombinedTrajectoryPoint(std::vector<double> &joint_point, const Eigen::Affine3d pose, double t, double tol,
-                          std::string name = std::string(), JointLockOptions options = JointLockOptions::LOCK_NONE,
-                          PointPreference type = PointPreference::JOINT, double timeout = 5.0)
+                          std::string name = std::string(), PointPreference type = PointPreference::JOINT,
+                          JointLockOptions options = JointLockOptions::LOCK_NONE, double timeout = 5.0)
     : TrajectoryPoint(name, t, (type == PointPreference::JOINT) ? PointType::JOINT : PointType::CARTESIAN, options)
     , options_(options)
     , joint_point_(joint_point)
--- a/moveit_simple/include/moveit_simple/point_types.h
+++ b/moveit_simple/include/moveit_simple/point_types.h
@@ -207,6 +207,8 @@ public:
     return timeout_;
   }
 
+  std::string pointVecToString(const std::vector<double> &vec) const;
+
 protected:
   
   virtual std::unique_ptr<JointTrajectoryPoint>
--- a/moveit_simple/include/moveit_simple/point_types.h
+++ b/moveit_simple/include/moveit_simple/point_types.h
@@ -81,7 +81,8 @@ protected:
   {
     UNKNOWN = 0,
     JOINT = 1,
-    CARTESIAN = 2
+    CARTESIAN = 2,
+    COMBINED = 3
   };
 
   TrajectoryPoint(std::string name, double t, PointType type, JointLockOptions options = JointLockOptions::LOCK_NONE)
@@ -102,6 +103,9 @@ protected:
   PointType type_;
   JointLockOptions joint_lock_options_ = JointLockOptions::LOCK_NONE;
 };
+
+
+
 
 class JointTrajectoryPoint : public TrajectoryPoint
 {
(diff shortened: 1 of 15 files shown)


moveit_simple/src/point_types.cpp
```diff
@@ -36,8 +35,9 @@ JointLockOptions TrajectoryPoint::getJointLockOptions()
   return joint_lock_options_;
 }
 
-std::unique_ptr<JointTrajectoryPoint> JointTrajectoryPoint::toJointTrajPoint(
-  const Robot &robot, double timeout, const std::vector<double> &seed, JointLockOptions options) const
+std::unique_ptr<JointTrajectoryPoint> JointTrajectoryPoint::toJointTrajPoint(const Robot &robot, double timeout,
```
shaun-edwards, line 38: Odd spacing?

moveit_simple/src/point_types.cpp
```diff
@@ -59,8 +59,9 @@ std::unique_ptr<CartTrajectoryPoint> JointTrajectoryPoint::toCartTrajPoint(const
   }
 }
 
-std::unique_ptr<JointTrajectoryPoint> CartTrajectoryPoint::toJointTrajPoint(
-  const Robot &robot, double timeout, const std::vector<double> &seed, JointLockOptions options) const
+std::unique_ptr<JointTrajectoryPoint> CartTrajectoryPoint::toJointTrajPoint(const Robot &robot, double timeout,
```
shaun-edwards, line 62: Odd spacing?

moveit_simple/src/point_types.cpp
```diff
@@ -82,4 +83,88 @@ std::unique_ptr<CartTrajectoryPoint> CartTrajectoryPoint::toCartTrajPoint(const
   return std::unique_ptr<CartTrajectoryPoint>(new CartTrajectoryPoint(*this));
 }
 
-} // namespace moveit_simple
+std::unique_ptr<JointTrajectoryPoint> CombinedTrajectoryPoint::toJointTrajPoint(const Robot &robot, double timeout,
```
shaun-edwards, line 86: Odd spacing

moveit_simple/src/point_types.cpp
```diff
@@ -82,4 +83,88 @@ std::unique_ptr<CartTrajectoryPoint> CartTrajectoryPoint::toCartTrajPoint(const
...
+
+  if (robot.getJointSolution(pose_, timeout, joint_point_, cart_point))
+  {
+    std::vector<double>::const_iterator joint_it = joint_point_.begin();
+    std::vector<double>::const_iterator cart_it = cart_point.begin();
+
+    while (in_tol && joint_it != joint_point_.end())
+    {
+      in_tol = std::abs(*joint_it - *cart_it) <= tol_;
+      joint_it++;
```
shaun-edwards: Is joint_point_ and cart_point guaranteed to be the same size?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
+uint8 NEW_WP=0				# Create a new waypoint at current location 			[required: waypoint name, optional: trajectory name]
```
shaun-edwards, line 1: This is a pretty complex service.  It's not likely that someone will create the backend code that support this service generically.  Given this, should we just open source our implementation?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
...
+uint8 DEL_WP=1				# Delete existing waypoint(s) from config 				[required: waypoint name(s)]
+uint8 GET_WP=2				# Request information on existing waypoint(s) 			[required: waypoint name(s)]
+uint8 UPDATE_WP=3			# Replace previous waypoint data with current location 	[required: waypoint name]	
+uint8 APPEND_WP_TO_TRAJ=4	# Add waypoint(s) to trajectory                         [required: trajectory name, waypoint name(s)] 
+uint8 REMOVE_WP_FROM_TRAJ=5 # Remove waypoint(s) from trajectory if it(they) exist 	[required: trajectory name, waypoint name(s)]
+uint8 NEW_TRAJ=10			# Create a new trajectory, optionally populate          [required: trajectory name, optional: waypoint name(s)]
+uint8 DEL_TRAJ=11			# Delete a trajectory from config						[required: trajectory name]
+uint8 GET_TRAJ=12			# Get all waypoints in a trajectory 					[required: trajectory name]
+uint8 CLEAR_TRAJ=13			# Clears the contents of a trajectory 					[required: trajectory name]
+uint8 RELOAD_YAML=255		# Update internal storage to match YAML                 [required: N/A]
```
shaun-edwards, line 11: How does the user know when to utilize the RELOAD method?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
...
+
+uint8 op_type
+string trajectory_name
+string[] waypoint_names
+---
+bool result
```
shaun-edwards, line 17: A `bool` result does not give us much information.  Should we define some result codes?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
+uint8 NEW_WP=0				# Create a new waypoint at current location 			[required: waypoint name, optional: trajectory name]
+uint8 DEL_WP=1				# Delete existing waypoint(s) from config 				[required: waypoint name(s)]
+uint8 GET_WP=2				# Request information on existing waypoint(s) 			[required: waypoint name(s)]
```
shaun-edwards, line 3: This is the only `op code` used in `moveit_simple`.  What are the others used for?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
...
+uint8 GET_TRAJ=12			# Get all waypoints in a trajectory 					[required: trajectory name]
+uint8 CLEAR_TRAJ=13			# Clears the contents of a trajectory 					[required: trajectory name]
+uint8 RELOAD_YAML=255		# Update internal storage to match YAML                 [required: N/A]
+
+uint8 op_type
+string trajectory_name
+string[] waypoint_names
+---
+bool result
+string message
```
shaun-edwards, line 18: What is the `message` member for?

moveit_simple/src/robot.cpp
```diff
@@ -377,8 +381,32 @@ std::unique_ptr<TrajectoryPoint> Robot::lookupTrajectoryPoint(const std::string&
   }
   else
   {
-    ROS_ERROR_STREAM("Failed to find point " << name << ", consider implementing more look ups");
-    throw std::invalid_argument("Failed to find point: " + name);
+    ROS_INFO_STREAM("Requesting trajectory training service info for waypoint: " << name);
+
+    std::vector<std::string> waypoint_names;
+    waypoint_names.push_back(name);
```
shaun-edwards: Is this an overload?  If we send a trajectory with a single waypoint, it only returns that waypoint?  Should we create another service?

moveit_simple_msgs/srv/TrajectoryTrainer.srv
```diff
@@ -0,0 +1,19 @@
...
+uint8 APPEND_WP_TO_TRAJ=4	# Add waypoint(s) to trajectory                         [required: trajectory name, waypoint name(s)] 
+uint8 REMOVE_WP_FROM_TRAJ=5 # Remove waypoint(s) from trajectory if it(they) exist 	[required: trajectory name, waypoint name(s)]
+uint8 NEW_TRAJ=10			# Create a new trajectory, optionally populate          [required: trajectory name, optional: waypoint name(s)]
+uint8 DEL_TRAJ=11			# Delete a trajectory from config						[required: trajectory name]
+uint8 GET_TRAJ=12			# Get all waypoints in a trajectory 					[required: trajectory name]
+uint8 CLEAR_TRAJ=13			# Clears the contents of a trajectory 					[required: trajectory name]
+uint8 RELOAD_YAML=255		# Update internal storage to match YAML                 [required: N/A]
+
+uint8 op_type
+string trajectory_name
```
shaun-edwards, line 14: Under what circumstances are these member variables filled out?  Should we consider creating multiple services?

moveit_simple/include/moveit_simple/point_types.h
```diff
@@ -154,5 +160,72 @@ class CartTrajectoryPoint : public TrajectoryPoint
...
+  virtual std::unique_ptr<JointTrajectoryPoint>
+  toJointTrajPoint(const Robot &robot, double timeout, const std::vector<double> &seed,
+                   JointLockOptions options = JointLockOptions::LOCK_NONE) const;
+
+  virtual std::unique_ptr<CartTrajectoryPoint> toCartTrajPoint(const Robot &robot) const;
+
+  bool compareJointAndCart(const Robot &robot, double timeout) const;
+
+private:
+
```
shaun-edwards, line 223: Why not utilize the Joint/Cart types here?  It seems like you are replicating the private variables of those types?

shaun-edwards (changes requested):

moveit_simple/include/moveit_simple/point_types.h
aaronplusone: To reduce redundant storage. Cart/Joint points have some of overlap, toJointTrajPoint and toCartTrajPoint already create points the sub-type.

moveit_simple/src/point_types.cpp
aaronplusone: That's what Clang Format gives me, but I can rearrange if it's non-standard.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
aaronplusone: Reload is intended for use when a user makes a change to the YAML file during runtime. By default, the YAML is only parsed on initialization, but I wanted the ability to update class members without killing/restarting the node.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
aaronplusone: Could do, but message is intended to provide diagnostic information when an operation is unsuccessful.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
aaronplusone: The others are used in creation and maintenance of trajectories. Use concept is that a robot can be manipulated (e.g. with the pendant) and its position can be stored through a service call. Trajectories are not in use in moveit_simple yet, but could be used in the future to retrieve sparse trajectories rather than adding points individually.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
aaronplusone: Message is intended to provide diagnostic information when an operation is unsuccessful. Generic name allows it to be used for user-specific purposes without too much confusion.

moveit_simple/src/robot.cpp
aaronplusone: GET_WP can return 1..n waypoints. It seemed unnecessary to create an additional case to ensure you only get a single waypoint. This call exists in the lookupTrajPoint call, but a lookupTrajectory could get a vector of waypoints. Same with lookupTrajPoints

moveit_simple_msgs/srv/TrajectoryTrainer.srv
aaronplusone: The formatting is off, but the comments specify what elements are necessary during the call. When performing trajectory manipulations, you need the trajectory name and sometimes waypoint names. Similar for waypoint manipulations. op_type is always required.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
shaun-edwards: But how does the user (in this case moveit_simple) know when to call the reload method.  I'm guessing another node edits the YAML, in which case, the moveit_simple node won't have any indication that is needs to update it's version.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
shaun-edwards: A human readable message is nice for logging, but how does the node know what to do?  Say the request fails because the file isn't loaded...we shouldn't be parsing the message string to programmatically respond.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
shaun-edwards: From a moveit_simple perspective, this is not utilized.  My general feeling is that this service is trying to be too much.  We can have a full featured set of services/op codes for nodes that edit the trajectory, but for those that simply read it, why not create a separate service.

moveit_simple/src/robot.cpp
shaun-edwards: A service should have an obvious sematic meaning.  Getting a waypoint and getting a trajectory (a set of waypoints) is obvious.  I suppose getting multiple waypoints is also a use case, but it might be confused with getting a trajectory.  I guess I could be convinced that we have services for all 3 use cases, but we should be careful about trying to combine all use cases into a single service.  It's not like services cost money.

moveit_simple_msgs/srv/TrajectoryTrainer.srv
gavanderhoorn: As an outsider observer: these sort of "union services" seem to go against the idea of having simple, explicit and clearly defined semantics. The interpretation of the fields in the request change (dramatically almost) based on the value of `op_type`. The type of the service itself does not carry enough semantic meaning to sufficiently explain the cases in which this can be used, or what to expect from an invocation of the service.

It might be worthwhile to provide additional services that implement some of this functionality and give them types that unambiguously convey their semantics. That is both user friendly *and* reduces the opportunity for mistakes when filling out fields (and the need the `op_type` field).

Finally: this might actually be a use-case for Actions, as some of these operations can potentially take a long time (fi: depending on how large the `.yaml` file is that is being loaded (long traj?) it could take going to disk, loading it, parsing, etc. All this time the caller is blocked.

aaronplusone: Alright, alright, I may have shoe-horned it a little bit. From the current MoveIt Simple perspective, I see get_waypoint (request a single waypoint) as critical for the lookupTrajectoryPoint extension. Additionally, I think get_waypoints (request multiple, possibly unrelated waypoints), and get_trajectory (request multiple, related waypoints) would be useful conveniences for building a trajectory.

Edit: These would be services tied here, and the others could be refactored to the training node.

As for using an action server, the extension to lookupTrajectoryPoint should block until a response regardless, no?

shaun-edwards (approved):

Issue #80, opened 2019-04-23
Robot Base class mutex


mlautman:
I am working to split the trajectory functionality out of the robot class and into it's own class. I noticed that the mutex strategy currently blocks this as the same mutex is used for the Robot functionality as is used by the Trajectory construction and planning. 

I am not sure I understand the mutex strategy here. 
1) Why does Robot, (a class that doesn't have any subscriptions) need a mutex?
2) Why doesn't the online robot class (which does have subscriptions) not use a mutex.

This seems like something we could address rather quickly but I wanted to check to see if I was missing anything before I invested time in a resolution

shaun-edwards:
>Why does Robot, (a class that doesn't have any subscriptions) need a mutex?

This is probably an artifact from when when Robot and Online Robot were the same class.

>Why doesn't the online robot class (which does have subscriptions) not use a mutex.

Again probably an oversight.  Does Online Robot require/assume a multi-threaded ROS spinner?

PR #82, opened 2019-04-24
limiting mutex locking to just protect changes in internal variables

mlautman:
This is related to issue #80 

It is unclear to me what the role of the mutex is in the Robot class but this limits the mutex to operations that update internal variables

@aaronplusone

### DIFF
--- a/moveit_simple/src/robot.cpp
+++ b/moveit_simple/src/robot.cpp
@@ -124,8 +124,6 @@ void Robot::addTrajPoint(const std::string& traj_name, const Eigen::Affine3d pos
                          double time, const InterpolationType& type, const unsigned int num_steps,
                          const std::string& point_name)
 {
-  std::lock_guard<std::recursive_mutex> guard(m_);
-
   ROS_INFO_STREAM("Attempting to add " << point_name << " to " << traj_name << "relative to" << frame << " at time "
                                        << time);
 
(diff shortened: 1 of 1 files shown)


mlautman:
@aaronplusone @henningkayser

moveit_simple/src/robot.cpp
```diff
@@ -124,8 +124,6 @@ void Robot::addTrajPoint(const std::string& traj_name, const Eigen::Affine3d pos
                          double time, const InterpolationType& type, const unsigned int num_steps,
                          const std::string& point_name)
 {
-  std::lock_guard<std::recursive_mutex> guard(m_);
```
henningkayser, line 127: I'm not sure that this is safe to remove since the calls to lookupTransform() can block the thread for up to 5 seconds so multiple threads could produce an indeterminate order of trajectory points. Probably a robot instance shouldn't be shared across multiple threads at all.

mlautman: This PR is intended to get @aaronplusone and @shaun-edwards's feedback and I don't believe that these changes are perfect as is. 

My belief is that we should only use mutex's when we are updating the internal data. If we need a mutex to protect the lookupTransform then the mutex should live in there. It seems strange to me that OnlineRobot does not have an equivalent mutex in it's updateCurrentState subscriber callback. We may need a bigger reassessment of the project's mutex strategy.

shaun-edwards: I'm hesitant to change any of these.  The recursive mutex is admittedly a sign of a poor implementation.  I don't believe we have any tests for the theadsafe-ness of moveit_simple.  Is this a must have change?

shaun-edwards (changes requested):

mlautman:
I think we can abandon this

PR #83, opened 2019-04-26
separate out simple type conversions

mlautman:
This should target a devel branch once it is made

This PR separates out simple conversions from the robot class. This is part of a larger effort to upgrade the trajectory planning capabilities.

### DIFF
--- a/moveit_simple/include/moveit_simple/conversions.h
+++ b/moveit_simple/include/moveit_simple/conversions.h
@@ -1,0 +1,46 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016 Shaun Edwards
+ * Copyright (c) 2018 Plus One Robotics
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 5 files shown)


mlautman:
@aaronplusone @henningkayser

moveit_simple/include/moveit_simple/conversions.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016 Shaun Edwards
```
henningkayser: This should be updated

moveit_simple/include/moveit_simple/conversions.h
```diff
@@ -0,0 +1,46 @@
...
+namespace moveit_simple
+{
+  /**
+   * @brief toJointTrajPtMsg - Converts native joint point (vector + time) to ROS joint
+   * trajectory point message type.
+   * @param joint_point
+   * @param time
+   * @return
+   */
+  trajectory_msgs::JointTrajectoryPoint toJointTrajPtMsg(const std::vector<double> &joint_point,
```
henningkayser, line 36: The point type implementation is object oriented so wouldn't this better fit as member functions, i.e. `toMsg()`, `setFromMsg()`?

moveit_simple/src/conversions.cpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016 Shaun Edwards
+ * Copyright (c) 2018 Plus One Robotics
```
henningkayser: update license

moveit_simple/include/moveit_simple/conversions.h
mlautman: Seems like scope creep. The intention here is to refactor methods out of Robot that don't need to be there. Wouldn't we need a new class that contains the vector of joints? It might be something to keep in mind while writing the trajectory planner class

shaun-edwards (approved):

PR #84, opened 2019-04-26
README improvements

mlautman:
This should target a master branch once that is created

this updates the README with some nice to have fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,35 @@ rosdep install --from-paths src --ignore-src --rosdistro $ROS_DISTRO -y
 4. Configure and Build
 ```
 cd $CATKIN_WS
-catkin config --extend /opt/ros/$ROS_DISTRO --cmake-args -DCMAKE_BUILD_TYPE=Release
+catkin init -w .
+catkin config --extend /opt/ros/$ROS_DISTRO \
+              --blacklist \
+                    motoman_bmda3_support \
+                    motoman_csda10f_moveit_config \
+                    motoman_csda10f_support \
+                    motoman_epx_support \
+                    motoman_experimental \
+                    motoman_gp12_support \
+                    motoman_gp7_support \
+                    motoman_gp8_support \
+                    motoman_mh12_support \
+                    motoman_mh50_support \
+                    motoman_motomini_support \
+                    motoman_mpl80_moveit_config \
+                    motoman_mpl_support \
+                    moveit_ros_planning_interface \
+                    moveit_experimental \
+                    moveit_planners_chomp \
+                    moveit_setup_assistant \
+                    moveit_ros_planning_interface \
+                    moveit_ros_manipulation \
+                    moveit_ros_visualization \
+                    motoman_driver \
+                    moveit_chomp_optimizer_adapter \
+                    moveit_ros_warehouse \
+                    moveit_planners_ompl \
+                    moveit_ros_perception \
+              --cmake-args -DCMAKE_BUILD_TYPE=Release
 catkin build
 ```
 **NOTE:** To use gdb with this project set `-DCMAKE_BUILD_TYPE=Debug` instead.
--- a/README.md
+++ b/README.md
@@ -3,24 +3,25 @@
 ## Installation
 1. Clone repository into workspace [It doesn't have to be called moveit_simple_ws]
 ```
-mkdir -p moveit_simple_ws/src && cd moveit_simple_ws/src
+export CATKIN_WS=$HOME/moveit_simple_ws/
+mkdir -p $CATKIN_WS/src && cd $CATKIN_WS/src
 git clone -b kinetic-devel https://github.com/plusone-robotics/moveit_simple.git
 ```
 
 2. Get Source Dependencies
 ```
-cd moveit_simple_ws/src
+cd $CATKIN_WS/src
 wstool init . moveit_simple/.travis.rosinstall
 ```
 
 3. Install Package Dependencies
 ```
-cd moveit_simple_ws
+cd $CATKIN_WS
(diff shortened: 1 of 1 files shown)


mlautman:
@aaronplusone @henningkayser

README.md
```diff
@@ -3,39 +3,65 @@
...
-mkdir src && cd src
-catkin_init_workspace
+mkdir -p moveit_simple_ws/src && cd moveit_simple_ws/src
 git clone -b kinetic-devel https://github.com/plusone-robotics/moveit_simple.git
 ```
 
-2. Source Dependencies
+2. Get Source Dependencies
 ```
+cd moveit_simple_ws/src
```
aaronplusone: You should already be in moveit_simple_ws/src from the first [command](https://github.com/plusone-robotics/moveit_simple/pull/84/files#diff-04c6e90faac2675aa89e2176d2eec7d8R6). This gives a "No such file or directory" error as written.

README.md
```diff
@@ -3,39 +3,65 @@
...
 wstool init . moveit_simple/.travis.rosinstall
-cd ..
 ```
 
-3. Package Dependencies
+3. Install Package Dependencies
 ```
+cd moveit_simple_ws
```
aaronplusone: Similar to above comment. If using relative path, this needs to be `cd ..`

README.md
```diff
@@ -3,39 +3,65 @@
...
 ```
-catkin build --cmake-args -DCMAKE_BUILD_TYPE=Debug
-or
 catkin build --cmake-args -DCMAKE_BUILD_TYPE=Release
+```
+**NOTE:** To use gdb with this project set `-DCMAKE_BUILD_TYPE=Debug` instead.
 
+5. Source the Workspace
+```
+cd moveit_simple_ws
```
aaronplusone: Same as above.

README.md
```diff
@@ -3,39 +3,65 @@
...
-## Running Tests with Rviz Visualization
+**NOTE:** You will need to re-run step #5 in any new terminal
+
+
+## Building and Running Tests
+
+1.  Run the MoveIt Simple Tests
+```
+roscd moveit_simple
+catkin run_tests --no-deps --this -i
```
aaronplusone: Can also be run from an arbitrary location in the workspace with `catkin build moveit_simple --no-deps -i --catkin-make-args run_tests`. Would eliminate the [roscd](https://github.com/plusone-robotics/moveit_simple/pull/84/files#diff-04c6e90faac2675aa89e2176d2eec7d8R41) above and [cd](https://github.com/plusone-robotics/moveit_simple/pull/84/files#diff-04c6e90faac2675aa89e2176d2eec7d8R47) below.

README.md
```diff
@@ -3,39 +3,65 @@
...
+
+### Run the Tests Using RViz as a Visualizer
+1. Launch the RViz visualizer
 ```
 roslaunch moveit_simple test_display.launch
+```
+
+2. Launch a Test. (Be sure to substitute `<Test Name>` with either `motoman_mh5` or `kuka_kr210`)
+```
 rostest moveit_simple <Test Name>_utest.launch -r
```
aaronplusone, line 94: Maybe obvious, but since we're touching this file already, it would be good to explicitly state that this instruction happens in a new terminal session.

aaronplusone (dismissed):

README.md
mlautman: good point. Fixed

README.md
mlautman: You are completely correct, these steps border on overly explicit. My personal preference is to write README's in a way where a user can copy-paste bash commands into their terminal from almost any state and it will run as expected. I made some changes that I think you will be happy with

mlautman:
@aaronplusone changes applied

README.md
```diff
@@ -3,39 +3,100 @@
...
-catkin_init_workspace
+export CATKIN_WS=$HOME/moveit_simple_ws/
+mkdir -p $CATKIN_WS/src && cd $CATKIN_WS/src
 git clone -b kinetic-devel https://github.com/plusone-robotics/moveit_simple.git
 ```
 
-2. Source Dependencies
+2. Get Source Dependencies
 ```
+cd $CATKIN_WS/src
```
aaronplusone: In my mind, having these `cd <path>` statements inhibit flow without a tangible benefit. Most people are going to copy and paste these instructions. If the desire here is to enforce you're in a specific directory when the command is run, you can make a compound instruction of `cd <path> && cmd`, but I think that too is unnecessary.

README.md
```diff
@@ -3,39 +3,100 @@
...
+cd $CATKIN_WS
 rosdep install --from-paths src --ignore-src --rosdistro $ROS_DISTRO -y
 ```
+4. Configure and Build
+```
+cd $CATKIN_WS
+catkin init -w .
+catkin config --extend /opt/ros/$ROS_DISTRO \
+              --blacklist \
+                    motoman_bmda3_support \
```
shaun-edwards, line 28: Out of curiosity, do we know where the `motoman_*` dependencies come from?

shaun-edwards (approved):
Just one question, but this is good to merge.

README.md
mlautman: They get downloaded in `motoman-experimental` which is in the rosinstall

README.md
mlautman: I don't see how these would inhibit flow as the new commands are even more copy-paste-able than the existing bash commands. The new README asks the user to set the CATKIN_WS environment variable whenever they `source devel/setup.bash` ensuring that whenever any of these commands are run, they get run in the correct context. 

I added the `&&`'s based on your request

PR #85, opened 2019-04-30
Simple demo

mlautman:
This PR adds a simple demo to the moveit_simple project so that developers can visually inspect the system after making changes without having to rebuild all of the tests which takes approximately 4 minutes. This PR builds off of the Readme changes in #84

PR #81, opened 2019-04-24
remove trailing whitespace

mlautman:
This PR removes the trailing whitespace across the code base. It was done with this command. 

```bash
find . \( -name '*.h' -o -name '*.hpp' -o -name '*.cpp' -o -name '*.c' -o -name '*.markdown' -o -name '*.md' -o -name '*.xml' -o -name '*.m' -o -name '*.txt' -o -name '*.sh' -o -name '*.launch' -o -name '*.world' -o -name '*.urdf' -o -name '*.xacro' -o -name '*.py' -o -name '*.cfg' -o -name '*.msg' -o -name '*.yml' -o -name '*.yaml' -o -name '*.rst' \) -exec sed -i 's/ *$//' '{}' ';'
```

PR #86, opened 2019-04-30
Trajectory planning

mlautman:
This separates the trajectory planning capabilities into it's own class. This accomplishes a few things. It reduces the size of the robot class to make it more manageable and sets the stage for a Descartes integration without adding a dependency to the main project. 

There are more methods that we could refactor out of the robot class and into the trajectory planner, but in an effort to keep this PR small, make as few API changes as possible and open the door for early feedback we think that this PR is ready for review as is.

This PR includes  #81, #83, #84,  and #85 which should be removed from this PR before merging

### DIFF
--- a/moveit_simple/include/moveit_simple/robot.h
+++ b/moveit_simple/include/moveit_simple/robot.h
@@ -34,6 +34,7 @@
 
 #include <moveit_simple/moveit_simple_dynamic_reconfigure_Parameters.h>
 #include <moveit_simple/point_types.h>
+#include <moveit_simple/trajectory_planner.h>
 
 namespace moveit_simple
 {
@@ -48,6 +49,8 @@ namespace moveit_simple
 class Robot
 {
 public:
+  friend class TrajectoryPlanner;
+
   /**
   * @brief Constructor
   */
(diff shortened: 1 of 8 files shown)


mlautman:
@aaronplusone @henningkayser

shaun-edwards (approved):

moveit_simple/src/robot.cpp
```diff
@@ -1083,177 +988,75 @@ bool Robot::jointInterpolation(const std::unique_ptr<TrajectoryPoint>& traj_poin
...
-    }
-  }
-
-  // Append the global points with local_points
-  points.insert(points.end(), points_local.begin() + 1, points_local.end());
-  ROS_INFO_STREAM("Appending trajectory point, size: " << points.size());
-  points_local.clear();
-  return true;
+  return trajectory_planner_.cartesianInterpolation(*this, traj_point, points, num_steps, collision_check);
+  // // Create a local vector for storing interpolated points and append
```
henningkayser: This should be removed

moveit_simple/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,323 @@
+/*
+ * Software License Agreement (Apache License)
+ *
+ * Copyright (c) 2016 Shaun Edwards
```
henningkayser: Update dates?

moveit_simple/include/moveit_simple/robot.h
```diff
@@ -538,6 +540,7 @@ class Robot
     const std::unique_ptr<JointTrajectoryPoint> &to, double t,
     std::unique_ptr<JointTrajectoryPoint> &point) const;
 
+  // TODO(mlautman): deprecate this
```
henningkayser: Why should this be deprecated? I thought this would just be passed to the TrajectoryPlanner

PR #87, opened 2019-04-30
Update MoveIt dependencies to current master branch

henningkayser:
This updates MoveIt dependencies so we can use all latest features.
The PR is still WIP and should later be merged to a new devel branch.


mlautman (approved):

moveit_simple/include/moveit_simple/point_types.h
```diff
@@ -140,13 +140,13 @@ class CartTrajectoryPoint : public TrajectoryPoint
   friend class CombinedTrajectoryPoint;
   CartTrajectoryPoint() : TrajectoryPoint("", 0.0, PointType::CARTESIAN) { }
 
-  CartTrajectoryPoint(const Eigen::Affine3d pose, const double t, std::string name = std::string(),
+  CartTrajectoryPoint(const Eigen::Isometry3d pose, const double t, std::string name = std::string(),
```
shaun-edwards, line 143: Just curious about the type change?  Is it an Eigen change or a more appropriate class?

shaun-edwards (changes requested):
See comment about changing Eigen types.  Will this be a API change?  Will others have to update their code to use the new type?

moveit_simple/include/moveit_simple/point_types.h
henningkayser: It's actually the more appropriate class. Affine3d allows shearing in the linear part so that `rotation()` implicitly computes SVD (https://github.com/ros-planning/moveit/issues/1073) which is unnecessarily costly. Isometry3d only consists of rotation and translation so it was changed in practically all ROS libraries since melodic and also in MoveIt master (https://github.com/ros-planning/moveit/pull/1096).

This change will however break compatibility with MoveIt kinetic. We could try to work around this by using templates and auto calls everywhere but I wouldn't recommend it.

Alternatively, we could also stick to Affine3d, apply changes introduced here https://github.com/ros-planning/moveit/pull/1091, and then only cast the transforms to Isometry3d by calling `matrix()` wherever necessary. For this we would have to guarantee that all transforms contain isometric matrices.

shaun-edwards: >This change will however break compatibility with MoveIt kinetic.

I think we already committed to this with this PR: https://github.com/plusone-robotics/moveit_simple/pull/88

henningkayser: > > This change will however break compatibility with MoveIt kinetic.
> 
> I think we already committed to this with this PR: #88

Yep, just showing the options.

mlautman: It boils down to the question, do we want to use moveit master branch with its fancy new features (TOTG, compatability with Ubuntu16 and 18 as well as ROS kinetic and melodic, etc..) or would we rather use the released Kinetic version of moveit and manually copy the desired features into the the moveit_simple package. Using the master branch seems like the right call long term but we would have to make a small API change and CI will take longer. 

You're call @shaun-edwards.

shaun-edwards: I'm comfortable making the upgrade.

shaun-edwards (approved):

PR #88, opened 2019-04-30
Trajectory validation and retiming waypoints

henningkayser:
This effort adds a check to the execute() function to validate all waypoints are reachable within the given time stamps and all values are within bounds. In case a trajectory is invalid an InvalidTrajectoryExecution is thrown. An optional parameter fix_trajectory should allow reparameterizing invalid trajectories.


henningkayser:
#92 fixes CI

shaun-edwards (approved):

PR #89, opened 2019-04-30
Limit joint windup for single IK solutions

henningkayser:
This is the first part on limiting joint windup for single IK solutions by setting a "soft" joint limit that still allows exceeding it intentionally. It's based on setting a joint value modifier in range [0,1] that is multiplied with the IK seed state to "pull" desired joints towards the center. This works best in combination with minimal-distance IK solvers (like currently LMA). To eliminate outliers completely the consistency limit of specified joints is set to the lower value PI.

This method allows to preserve the full freedom of the joint since only the seed state is modified.
This allows for consecutive IK solutions to still move the full way if the steps are sufficiently small (i.e. for cartesian planning). For example a value of 0.5 would allow the joint value to converge to 2PI but only if sequential IK calls produce solutions within closest proximity. The closer the the value is to the limit the lower the distance needs to be to still move towards it. In practice this means that the actually reachable limit is considerably below that one. Lower fractions let the value converge to lower limits (I could make a table for that). A good value to limit wrist windup would be 0.3 or maybe even lower (~1.4PI limit).

The modifiers are stored in a map ik_seed_state_fractions_ and can only accessed by getter() and setter() functions that also test the validity of the values. The feature is enabled by calling robot.limitJointWindup(true) and is active for all getJointSolution() calls including waypoint conversions.

The test shows how this is applied to limit wrist joint windup by setting the corresponding fraction to 0.3.
I suggest specifying desired values either inside the kinematics.yaml or inside a new moveit_simple.yaml since I think this should be needed anyway.

This solution is meant to be used for all IK calls but does not support end effector symmetries.
For this a separate function will be needed anyway just as for separate pick/place state optimization.
I'm planning to do follow-up PRs for that so this one doesn't get bloated.

### DIFF
--- a/moveit_simple/include/moveit_simple/robot.h
+++ b/moveit_simple/include/moveit_simple/robot.h
@@ -435,6 +435,12 @@ public:
    */
   double getSpeedModifier(void) const;
 
+  void limitJointWindup(bool enabled);
+
+  bool setIKSeedStateFractions(const std::map<size_t, double>& ik_seed_state_fractions);
+
+  std::map<size_t, double> getIKSeedStateFractions() const;
+
   void updateRvizRobotState(const Eigen::Isometry3d &pose, const std::string &in_frame,
     const std::string &joint_seed, double timeout = 10.0) const;
 
@@ -557,10 +563,10 @@ protected:
   virtual std::vector<double> getJointState(void) const;
 
   bool getIK(const Eigen::Isometry3d pose, const std::vector<double> &seed,
-    std::vector<double> &joint_point, double timeout = 1) const;
+    std::vector<double> &joint_point, double timeout = 1, bool limit_joint_windup = false) const;
 
   bool getIK(const Eigen::Isometry3d pose, std::vector<double> &joint_point,
-    double timeout = 1) const;
+    double timeout = 1, bool limit_joint_windup = false) const;
 
   bool getFK(const std::vector<double> &joint_point, Eigen::Isometry3d &pose) const;
 
@@ -613,6 +619,10 @@ protected:
 
   std::string planning_group_;
   std::string robot_description_;
+
+  // Limit IK joint windup
+  bool limit_joint_windup_;
+  std::map<size_t, double> ik_seed_state_fractions_;
 };
 } // namespace moveit_simple
 #endif // ROBOT_H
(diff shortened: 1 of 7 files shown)


henningkayser:
#92 fixes CI

moveit_simple/src/robot.cpp
```diff
@@ -1240,16 +1231,35 @@ bool Robot::getFK(const std::vector<double>& joint_point, Eigen::Isometry3d& pos
...
+  return getIK(pose, joint_point, timeout, limit_joint_windup);
 }
 
-bool Robot::getIK(const Eigen::Isometry3d pose, std::vector<double>& joint_point, double timeout) const
+bool Robot::getIK(const Eigen::Isometry3d pose, std::vector<double>& joint_point, double timeout,
+                  bool limit_joint_windup) const
 {
   Eigen::Isometry3d ik_tip_pose = pose * ik_tip_to_srdf_tip_;
-  if (virtual_robot_state_->setFromIK(joint_group_, ik_tip_pose, timeout))
+  std::vector<double> consistency_limit(6, 4 * M_PI);
```
mlautman, line 1254: You should not be hardcoding magic numbers here.

moveit_simple/include/moveit_simple/robot.h
```diff
@@ -557,10 +583,10 @@ class Robot
   virtual std::vector<double> getJointState(void) const;
 
   bool getIK(const Eigen::Isometry3d pose, const std::vector<double> &seed,
-    std::vector<double> &joint_point, double timeout = 1) const;
+    std::vector<double> &joint_point, double timeout = 1, bool limit_joint_windup = false) const;
```
mlautman: If this is already a class variable you shouldn't need to pass it in here.

henningkayser: @mlautman this was intended since it's crucial to allow `getIK()` calls with and without joint windup limitation. The class variable should be used for the public interface functions (i.e. getJointSolution()) and `getIK()` is a private helper where this feature should be optional independent of the current setting. For instance `getPickPlaceJointSolutions()` in #91 allows using joint windup limitation for pick pose IK but then must call `getIK()` with limit_joint_windup=false so that the object orientation is not invalidated.

PR #90, opened 2019-04-30
Support IK solutions with end effector symmetries

henningkayser:
As discussed in https://github.com/plusone-robotics/moveit_simple/issues/79 pick and place poses should be computed such that tool windup (or joint windup in general) is reduced as much as possible.
With symmetric end effectors (circular, rectangular, square) it's possible to rotate the grasp pose while still producing valid grasps. This extended IK wrapper function adds support for computing IK solutions for all three end effector types and also optionally applies the tool windup reduction feature introduced in #89.

PR #91, opened 2019-05-01
Simple combined pick/place IK supporting symmetries

henningkayser:
Based on #90 this PR adds a simple function `getPickPlaceJointSolutions()` that computes IK solutions for pick and place poses while optimizing for symmetric end effectors. This first implementation uses `getSymmetricIK()` for searching the grasp state and then uses the orientation offset to compute the place joint state via normal IK.


henningkayser:
This PR adds a71dad8 and e7db1c5 .
I see this as a preparation for optimized implementations.

henningkayser:
#92 fixes CI

PR #93, opened 2019-05-31
Check to make sure that the goal has been reached

kriskozak:
- Compare current state to goal state in execute() to make sure
  that the goal has actually been reached. This provides some
  extra assurance that that a failure to reach the goal will be
  detected. This addresses an issue that was identified with at
  least one robot/driver.
- Closeness to the goal is set by a reconfigurable parameter that
  was already present in robot/online_robot, and is used for an
  equivalent purpose.
- Add more detail to the log messages.

### DIFF
--- a/moveit_simple/src/online_robot.cpp
+++ b/moveit_simple/src/online_robot.cpp
@@ -127,13 +127,37 @@ void OnlineRobot::execute(std::vector<moveit_simple::JointTrajectoryPoint> &join
 
     ros::Duration timeout(TIMEOUT_SCALE * traj_time.toSec());
 
-    if (action_.sendGoalAndWait(goal, timeout) == actionlib::SimpleClientGoalState::SUCCEEDED)
+    actionlib::SimpleClientGoalState result = action_.sendGoalAndWait(goal, timeout);
+    if (result == actionlib::SimpleClientGoalState::SUCCEEDED)
     {
-      ROS_INFO_STREAM("Successfully executed Joint Trajectory ");
+      std::vector<double> joint_state = getJointState();
+      std::vector<double> joint_state_goal = joint_trajectory_points.back().jointPoint();
+
+      bool reached_goal = true;
+      for (size_t i = 0; i < joint_state.size(); ++i)
+      {
+        // Use the dynamic param joint_equality_tolerance_ to check whether we
+        // are close enough to the goal in joint space.
+	if (std::abs(joint_state[i] - joint_state_goal[i]) > joint_equality_tolerance_)
+        {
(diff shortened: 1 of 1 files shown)


shaun-edwards:
+1 from me.

shaun-edwards:
@kriskozak, CI is failing now.  It wasn't failing in the past.  Have you verified that the tests passed on your local machine?

kriskozak:
I can't seem to reproduce the unit test failure that is in the ci log.  Here's where the test seems to fail on the CI server:

```
[moveit_simple.rosunit-kuka_kr210_utest/speed_reconfiguration][FAILURE]---------
/root/catkin_ws/src/moveit_simple/moveit_simple/test/kuka_kr210.cpp:468
Expected: robot->execute(TRAJECTORY_NAME) doesn't throw an exception.
  Actual: it throws.
```

When I run the tests locally, I see a failure (not every time, though) that is different, but I see it on both the PR branch and the target kinetic-devel branch.  Here is where it fails locally:

```
[moveit_simple.rosunit-motoman_mh5_utest/speed_reconfiguration][FAILURE]--------
/kozak_ws/src/moveit_simple/moveit_simple/test/motoman_mh5.cpp:588
The difference between delta_time_for_speed_limits and 0.0 is 0.29256852822645385, which exceeds execution_time_tolerance, where
delta_time_for_speed_limits evaluates to -0.29256852822645385,
0.0 evaluates to 0, and
execution_time_tolerance evaluates to 0.25.
```
Note that the CI failure was for the kuka test, while the local failure was for the motoman test. Both failed the "speed reconfiguration" test, but I'm not sure whether that is a coincidence.

Actually, looking back through the logs, the kuka test did fail locally once (but that was before I got all the test dependencies installed, so other tests were also failing)...

After running it one more time I saw the kuka test fail in the same way as the motoman test:
```
[moveit_simple.rosunit-kuka_kr210_utest/speed_reconfiguration][FAILURE]---------
/kozak_ws/src/moveit_simple/moveit_simple/test/kuka_kr210.cpp:483
The difference between delta_time_for_speed_limits and 0.0 is 0.25862099645159042, which exceeds execution_time_tolerance, where
delta_time_for_speed_limits evaluates to -0.25862099645159042,
0.0 evaluates to 0, and
execution_time_tolerance evaluates to 0.25.
```

Given the fact that I'm not able to reliably reproduce the failure that was experienced on the CI, and I am seeing a possibly related error pop show up on both branches, I'm guessing that this may just be the result of one or more unreliable unit tests.

Perhaps try re-running it and see if it fails the same way again.

